Issue #47, opened 2026-08-05
Sometimes delete doesn't remove the loop from sidebar, killing the app and restarting removes it


scgopi:

scgopi:
[graphcode-fixer] Classifying this as a **bug** — a delete that takes effect in the daemon but leaves its row on screen is incorrect behaviour, and it's in scope. Labelled `bug`; a fix is up as a PR.

**Root cause.** Not in the delete path at all — the delete always worked. `graphcoded` keys a client's joined projects by *connection* (`ProjectRegistry.connectionProjectPaths`), and the app asks to join exactly once, from `AppFeature.task` (`.restoreOpenProjects` + `.openGlobalGraph`). `DaemonConnection.events()` reconnects on any I/O failure — a daemon restart, a frame that failed to decode — but never re-announces those projects. The replacement socket is therefore attached to no `GraphStore` at all. Commands still travel *up* it fine (`ProjectRegistry` routes `.graphCommand` by looking the store up in `stores`, which doesn't care who is connected), so the loop really is deleted and really is persisted — but the confirming `.graphChanged` broadcast goes to the connection that had gone, and the sidebar keeps rendering the graph it last heard about.

That accounts for all three parts of the report: it's intermittent because it only bites once a reconnect has happened; the loop is genuinely gone daemon-side; and relaunching fixes it because a fresh launch re-sends the join commands.

Worth knowing: while the app is in that state, *every* graph update is missing, not just deletes — new loops, renames, and state changes are all equally stale. Delete is simply the one where a stale row is unmistakably wrong.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #33, opened 2026-08-04
Check for Updates: offer beta releases to beta builds (update channel)


scgopi:
Check for Updates (from #27) asks `releases/latest`, which excludes prereleases — correct for stable installs, but a beta build (e.g. 0.1.15-beta2) never hears about newer betas.

Add beta-channel support: a build whose installed version is a prerelease (or an explicit setting) should also consider prereleases when checking — query the releases list, pick the highest version by the existing `AppUpdate` ordering (which already parses `betaN` suffixes), and offer it the same way. A stable install with the setting off should behave exactly as today.

Acceptance: an installed 0.1.15-beta1 sees 0.1.15-beta2 and is walked to its DMG; a stable install is still only offered stables by default.

scgopi:
For the record (a peer loop asked): this shipped in **v0.1.15** via PR #34 — channel derived from the installed version, releases-list query on the beta channel, closed 2026-08-04T02:31Z when #34 merged. v0.1.16's #36/#37 (PR #38) built *on top of* this: the Settings switch writes the same `updateChannel` default this issue introduced, and in-place install replaces the browser walk for whatever the channel offers. Nothing from this issue is missing or superseded.

Issue #46, opened 2026-08-05
I see a bug when the machine in graphcode is restarted. Relaunching  graphcode brings loops starting freshly instead of attaching to the existing copilot/claude sessions. Store it somewhere to re-attach to the same session


scgopi:

scgopi:
[graphcode-fixer] Classifying this as **ambiguous — provisionally out of scope for an automated fix**, because it reads as either a broken guarantee (bug) or a request for new session-resume persistence (enhancement), and the two have opposite fixes.

**One question that settles it: did the Mac itself reboot, or did you only quit and relaunch GraphCode with the machine still up?**

Why it matters:

- **If the machine rebooted** — this is working as designed, and what you're asking for is a new capability. A loop's terminal is a `zmx` session, and re-attach works only because the zmx server process holds the PTY. A reboot kills that process, so there is nothing left to attach to. `graphcoded` then runs `ensureUnattendedSessions()`, finds no live session, and relaunches each unattended loop fresh with its opening prompt — exactly the "loops starting freshly" you saw. The documented guarantee is deliberately scoped to *closing the window, quitting the app, and restarting the daemon* (`docs/index.md:114`, `README.md:48`); a reboot is not in that list. "Store it somewhere to re-attach" would mean persisting each backend's own session id and resuming with `claude --resume` / the Copilot equivalent — which the code currently rules out on purpose (`GraphcodeKit/Sources/Sessions/AgentEnvironment.swift:10`: "no session id, no `/resume` later"). That's an enhancement, and a fairly substantial one.

- **If the machine stayed up and you only relaunched the app** — that is a real bug, since surviving an app quit *is* a documented guarantee. Session names are `graphcode-<node UUID>` and layouts are persisted per node, so a fresh session on relaunch would mean node ids or the saved layout are not surviving as intended, and I'd chase that down.

I have not labelled this yet, to avoid mislabelling it either way. Answer with which of the two happened and the next run will pick it up — as a fix if it's the second case.

---
_Generated by [Claude Code](https://claude.ai/code)_

scgopi:
[...]

PR #50, opened 2026-08-05
0.1.21-beta1 (build 65): version bump for the daemon-reconnect fix (#48)

scgopi:
Version bump only — this is the repo-side half of cutting `0.1.21-beta1`. **The build itself cannot be produced in the cloud sandbox and still has to run on your Mac** (see the checklist below).

## What this carries

`main` currently sits at `f275c57`, which merged #48 (`fix: sidebar goes stale after a daemon reconnect, so a deleted loop keeps its row`, issue #47). That fix is on `main` but is not in any release — `Project.swift` still shipped `0.1.20` / build 64, i.e. the v0.1.20 stable. This bump is what makes the next beta contain it.

## Change

`Project.swift` only, two lines:

- `CFBundleShortVersionString`: `0.1.20` → `0.1.21-beta1`
- `CFBundleVersion`: `64` → `65`

Following the convention recorded in the surrounding comment and in `0.1.19`/61 → `0.1.20-beta1`/62: since 0.1.15-beta3 a beta carries its **full suffixed version** in `CFBundleShortVersionString`, which is how Check for Updates knows the install is on the beta channel (#33). The build number increments monotonically.

## Verified in cloud

- `grep -rn "0\.1\.20"` across the repo (excluding `ThirdParty/` and `docs/`) returns exactly one hit — `Project.swift:57`. There is no second place carrying the version, so this is the whole bump.
- A beta's version is deliberately **not** read out of `Project.swift` by `make tap-bump`; that channel requires an explicit `VERSION=`, so nothing else needs editing for the tap.
- Diff re-read line by line: two string literals, no structural change, comments untouched.

## NOT verified — needs your Mac

Nothing here was compiled or built. There is no Xcode, macOS SDK, signing certificate, or notarization credential in this sandbox.

- [ ] `make generate` — Tuist regenerates the project with the new Info.plist values
- [ ] `make check` / `make test`
- [ ] `make build-app`, confirm About reports `0.1.21-beta1` and build `65`
- [ ] `make release-dmg SIGN_ID=... NOTARIZE=1`
- [ ] `gh release create 0.1.21-beta1` (beta tags carry **no** `v` prefix)
- [ ] `make tap-bump CHANNEL=beta VERSION=0.1.21-beta1` — run *after* the release exists, since the checksum is read back from the served asset
- [ ] Manual repro of #47 against the beta: open a project, force a daemon reconnect, delete a loop, confirm the sidebar row goes with it

---
_Generated by [Claude Code](https://claude.ai/code/session_01CRtNLBrHUBWCPhxLKH1EKq)_

### DIFF
--- a/Project.swift
+++ b/Project.swift
@@ -54,8 +54,8 @@ let project = Project(
                 // (#33). Before that the suffix lived on the tag only, so betas 48/49
                 // of the 0.1.15 line read "0.1.15" and are told by the build number
                 // apart.
-                "CFBundleShortVersionString": "0.1.20",
-                "CFBundleVersion": "64",
+                "CFBundleShortVersionString": "0.1.21-beta1",
+                "CFBundleVersion": "65",
             ]),
             resources: [
                 "graphcode/Resources/**"


scgopi:
Superseded by #53 which includes this version bump along with fixes for #51, #52, #46.

PR #72, opened 2026-08-08
fix: resume remote loops after the remote host reboots

scgopi:
## The issue

When a **remote** machine restarts — a Codespace idle-stopping and starting again — its loops die and nothing brings them back. Relaunching the app eventually restarts them **from their opening prompt**, redoing work. This is exactly what `SessionIDStore`'s own doc comment says it exists to prevent.

A *local* reboot was handled end to end: the machine comes back → `graphcoded` restarts → loading the persisted graph calls `ensureUnattendedSessions` → each loop resumes from the session ID a `SessionStart` hook wrote. Every link in that chain turned out to be local-only.

## Four breaks, all remote-only

| # | Break | Was |
|---|---|---|
| 1 | **Nothing re-ensured.** `ensureUnattendedSessions` runs once per daemon lifetime, at project load. A remote host reboots without this daemon restarting, so it never fired again. | `ProjectRegistry.swift:275` |
| 2 | **`start()` short-circuited before the resume logic.** The remote early-return happens before `SessionIDStore` is consulted, so the remote branch was always the from-scratch launch. | `ZmxSessionLauncher.swift:988-991` |
| 3 | **`resumeArguments` refused remote outright** — `guard remote == nil else { return nil }`, present since the feature landed in `faca382`. | `ZmxSessionLauncher.swift:597-598` |
| 4 | **The session ID was never captured on the host anyway.** `remoteWriteFragment` reused the local hooks JSON, which bakes in *this Mac's* `/Users/<me>/.graphcode`. On a Codespace that `mkdir -p` fails as a non-root user, so nothing was ever written. | `PresenceHooks.swift:92` |

## The fix

**A — decide resume-or-fresh in the remote shell.** `resumeArguments` now takes the same three branches `arguments(forNode:)` already takes for remote: no local hooks file, no local `zmx` to report to, `--settings` pointed at the host's own copy. `remoteEnsureInvocation` gained a `remoteCreateScript` fallback inside the *existing* single round-trip:

```sh
zmx get <name> >/dev/null 2>&1 || {
  GRAPHCODE_RESUME_ID=$(cat "$HOME/.graphcode/sessions/<uuid>.id" 2>/dev/null)
  if [ -n "$GRAPHCODE_RESUME_ID" ]; then <zmx run … --resume "$GRAPHCODE_RESUME_ID">
  else <zmx run … with the opening prompt>; fi
}
```

The choice has to be made there: `SessionIDStore.load` reads this Mac's disk, where a remote loop's ID has never existed. A sentinel argv element (`remoteResumeIDPlaceholder`) carries the ID through argv assembly, and `remoteQuotedCommand` is the one place that renders it unquoted — every other argument stays a shell-quoted literal.

**B — capture the ID where it will be read.** `captureSessionID` now takes the sessions directory as shell *text* rather than a path, so the remote hook names `"$HOME/.graphcode/sessions"` — the same reason `remotePathExpression` was already an expression. Local callers pass the absolute path and are byte-identical to before.

**C — notice the reboot at all.** A new 60s `ProjectRegistry` liveness sweep calls `GraphStore.ensureUnattendedSessionsAlive()` for remote-project stores only. Unlike the presence poll it runs with **no client attached**, because a loop being alive matters when nobody is watching and a reading does not. Two deliberate differences from the load-time ensure, both because this repeats:

- **No goal-poller re-arming** — `armGoalPoller` replaces the existing poller, so re-arming every sweep would restart its interval and a goal polled less often than the sweep would never fire.
- **Resolved nodes skipped whatever their loop type** — the load-time version restarts a `.stopped` time-based node, defensible once at boot and wrong every minute. A loop a human stopped stays stopped.

Without C, A and B only help when the *Mac's* daemon restarts.

## Codespaces caveat

`--resume` needs the transcript in `~/.claude/projects/`. A **stop/start** preserves the container filesystem, so resume works. A **rebuild** wipes everything outside `/workspaces` — the `-n` test then falls through to a fresh launch, which is the correct fallback rather than a hang.

## Local loops

Every touched path branches on `RemoteProjectLocation.parse(...) != nil` and takes its previous shape when that is `nil`:

- `resumeArguments` with `remote == nil` computes the same `hooksFile`, `reportingPath`, and an empty `scriptSuffix` — identical argv.
- `PresenceHooks.json(forBackend:zmxPath:)` without the new parameter emits the same string; `localSessionsExpression` resolves to the same `SessionIDStore` directory.
- The placeholder carve-out in `remoteQuotedCommand` can only match a sentinel no real value can be.
- The sweep filters on remote project paths, so local stores and the `graphcode://global` graph are never touched.

## Tests

New `RemoteSessionResumeTests` (11 cases): the remote hook writing into the remote `$HOME`, the ensure dial reading it back, `--resume` riding as a variable reference, the fresh-launch fallback, the existence check still guarding both branches, the sweep restarting unattended loops without re-arming pollers, the sweep leaving a stopped loop stopped — plus regression guards that the local hook still names the local store and a local resume still carries its literal ID with no `$HOME` in the argv.

Full suite: **723 passing**, exit 0. `swiftlint` and `swift format lint` show no new violations (the remaining ones pre-exist on `main`).

### DIFF
--- a/GraphcodeKit/Sources/GraphStore.swift
+++ b/GraphcodeKit/Sources/GraphStore.swift
@@ -1380,6 +1380,27 @@ public actor GraphStore {
     }
   }
 
+  /// The session half of `ensureUnattendedSessions`, for the repeating remote liveness
+  /// sweep (`ProjectRegistry.startRemoteLivenessSweep`). A remote host can reboot while
+  /// this daemon keeps running, and nothing else notices: the store was loaded long ago,
+  /// so the one call site that restarts loops after a reboot is the *local* machine's.
+  ///
+  /// Two deliberate differences from the load-time version, both because this runs on a
+  /// timer rather than once:
+  ///
+  /// - **No goal pollers.** `armGoalPoller` replaces the existing one, so re-arming every
+  ///   sweep would restart each poller's interval and a goal polled less often than the
+  ///   sweep would never fire at all. The pollers are already running; they are in-memory
+  ///   and a remote reboot doesn't touch them.
+  /// - **Resolved nodes are skipped whatever their loop type.** The load-time version
+  ///   restarts a `.stopped` time-based node, which is defensible once at boot and wrong
+  ///   every minute: a human who stopped a remote loop would watch it come back.
+  public func ensureUnattendedSessionsAlive() {
+    for node in graph.nodes where node.runsUnattended && !node.isResolved {
+      ensureSession(node)
+    }
+  }
+
   // MARK: - Broadcast
 
   private func broadcast() {
--- a/GraphcodeKit/Sources/ProjectRegistry.swift
+++ b/GraphcodeKit/Sources/ProjectRegistry.swift
@@ -118,6 +118,47 @@ public actor ProjectRegistry {
     presencePoller = nil
   }
 
+  // MARK: - Remote liveness
+
+  /// How often a remote project's unattended sessions are checked for existence.
+  ///
+  /// Local loops need no such sweep: their sessions die with the machine, and the machine
+  /// coming back restarts `graphcoded`, which loads the graph and calls
+  /// `ensureUnattendedSessions`. A *remote* host reboots — a Codespace stops on idle and
+  /// starts again — without this daemon restarting at all, so that one call site never
+  /// fires and every loop on that host stays dead until the app is relaunched.
+  ///
+  /// Unlike the presence poll this runs with no client attached, because a loop being
+  /// alive matters when nobody is watching and a reading does not. The cost is one ssh
+  /// per unattended remote loop per minute, multiplexed onto the host's existing
+  /// `ControlMaster` connection — a quarter of what the presence poll already spends per
+  /// loop while the app is open.
+  static let remoteLivenessSweepInterval: Duration = .seconds(60)
+
+  private var remoteSweeper: Task<Void, Never>?
+
+  /// Started by the first remote project this daemon loads and left running: a store is
+  /// never removed for being idle, and the sweep costs nothing on a tick where no project
+  /// is remote.
+  private func startRemoteLivenessSweep() {
+    guard remoteSweeper == nil, ensureSession != nil else { return }
+    remoteSweeper = Task { [weak self] in
+      while !Task.isCancelled {
+        try? await Task.sleep(for: Self.remoteLivenessSweepInterval)
+        guard !Task.isCancelled else { return }
+        await self?.sweepRemoteSessions()
+      }
+    }
+  }
+
+  /// Sequential across projects for the same reason `pollPresence` is: one burst of ssh
+  /// dials per tick is what the multiplexed connection exists to avoid.
+  private func sweepRemoteSessions() async {
+    for (path, store) in stores where RemoteProjectLocation.parse(projectPath: path) != nil {
+      await store.ensureUnattendedSessionsAlive()
+    }
+  }
+
   /// Sequential rather than concurrent across projects: each store's poll already spawns
   /// one subprocess per loop, and firing every project's at once would turn a quiet
   /// background tick into a burst of them.
@@ -273,6 +314,9 @@ public actor ProjectRegistry {
     // but not a reboot, so something has to restart it, and this is the moment the
     // persisted graph is first seen. Already-running sessions are left untouched.
     await newStore.ensureUnattendedSessions()
+    // The load-time ensure above is the *local* machine's reboot recovery. A remote host
+    // reboots on its own schedule, so its loops need a repeating check as well.
+    if RemoteProjectLocation.parse(projectPath: path) != nil { startRemoteLivenessSweep() }
     return newStore
   }
 
--- a/GraphcodeKit/Sources/ProjectRegistry.swift
+++ b/GraphcodeKit/Sources/ProjectRegistry.swift
@@ -129,10 +129,10 @@ public actor ProjectRegistry {
   /// fires and every loop on that host stays dead until the app is relaunched.
   ///
   /// Unlike the presence poll this runs with no client attached, because a loop being
-  /// alive matters when nobody is watching and a reading does not. The cost is one ssh
-  /// per unattended remote loop per minute, multiplexed onto the host's existing
-  /// `ControlMaster` connection — a quarter of what the presence poll already spends per
-  /// loop while the app is open.
+  /// alive matters when nobody is watching and a reading does not. On a healthy tick the
+  /// dial is a bare `zmx get` — `remoteEnsureInvocation` keeps the hooks write and the
+  /// file delivery behind that check precisely so this can be cheap — multiplexed onto
+  /// the host's existing `ControlMaster` connection.
   static let remoteLivenessSweepInterval: Duration = .seconds(60)
 
   private var remoteSweeper: Task<Void, Never>?
@@ -151,8 +151,16 @@ public actor ProjectRegistry {
     }
   }
 
-  /// Sequential across projects for the same reason `pollPresence` is: one burst of ssh
-  /// dials per tick is what the multiplexed connection exists to avoid.
+  /// The registry outlives the daemon in practice, but a `Task` holding only a weak
+  /// `self` would otherwise keep waking every minute after the registry it sweeps for is
+  /// gone — the same reason `GraphStore` cancels its goal pollers here.
+  deinit { remoteSweeper?.cancel() }
+
+  /// `ensureSession` is fire-and-forget by contract (`CLISessionBackend.ensureSession`
+  /// spawns a detached task), so this returns well before the dials it started finish and
+  /// the loop below is an ordering, not a throttle. What bounds the work is
+  /// `RemoteEnsureGate`: one ensure per node at a time, so a slow tick cannot pile a
+  /// second dial onto the same session.
   private func sweepRemoteSessions() async {
     for (path, store) in stores where RemoteProjectLocation.parse(projectPath: path) != nil {
       await store.ensureUnattendedSessionsAlive()
--- a/GraphcodeKit/Sources/Sessions/PresenceHooks.swift
+++ b/GraphcodeKit/Sources/Sessions/PresenceHooks.swift
@@ -75,25 +75,50 @@ public enum PresenceHooks {
   /// (`graphcode-<UUID>`), and the ID is written to the `SessionIDStore` path. Silently
   /// skipped when either variable is absent — a session graphcode didn't start, or a
   /// Claude Code version that doesn't export its session ID.
-  static func captureSessionID(supportDir: String) -> String {
-    let dir = "\(supportDir)/sessions"
-    return "if [ -n \"$ZMX_SESSION\" ] && [ -n \"$CLAUDE_CODE_SESSION_ID\" ]; then "
-      + "node_id=\"${ZMX_SESSION#graphcode-}\"; "
-      + "mkdir -p \(singleQuoted(dir)); "
-      + "printf '%s' \"$CLAUDE_CODE_SESSION_ID\" > \(singleQuoted(dir))/\"$node_id\".id; "
+  ///
+  /// `sessionsDirectory` is shell *text*, not a path, for the same reason
+  /// `remotePathExpression` is one: a hook that runs on another machine has to name that
+  /// machine's home directory, and only a shell there can expand it. A hook file carrying
+  /// this Mac's `/Users/<me>/.graphcode` wrote nothing at all on a Codespace, where
+  /// `/Users` doesn't exist and the agent user can't create it.
+  static func captureSessionID(sessionsDirectory: String) -> String {
+    "if [ -n \"$ZMX_SESSION\" ] && [ -n \"$CLAUDE_CODE_SESSION_ID\" ]; then "
+      + "node_id=\"${ZMX_SESSION#\(SurfaceRef.zmxSessionPrefix)}\"; "
+      + "mkdir -p \(sessionsDirectory); "
+      + "printf '%s' \"$CLAUDE_CODE_SESSION_ID\" > \(sessionsDirectory)/\"$node_id\".id; "
       + "fi; exit 0"
+  }
+
+  /// Where `captureSessionID` writes, as the shell text each side needs: a quoted
+  /// absolute path locally (the `SessionIDStore` directory), a `$HOME` expression
+  /// remotely.
+  static var localSessionsExpression: String {
+    singleQuoted(SupportDirectory.url.appendingPathComponent("sessions", isDirectory: true).path)
+  }
+
+  public static let remoteSessionsExpression = "\"$HOME/.graphcode/sessions\""
+
+  /// The remote twin of `SessionIDStore.file(forNodeID:)` — the file the remote
+  /// `SessionStart` hook wrote, as a shell expression the ensure dial can `cat`. The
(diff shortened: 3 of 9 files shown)


scgopi:
## Independent review pass — 5 defects found and fixed (`20369b5`)

An adversarial review was run against `8fc9f89`, focused on whether any of this regresses local folders and local loops.

**Local regression verdict: no.** Evidence checked:

| Area | Finding |
|---|---|
| `localSessionsExpression` vs `main` | Identical string across 5 `SupportDirectory.url` cases (default, relative `GRAPHCODE_SUPPORT_DIR`, absolute override with trailing slash, `~`-prefixed override, `/` as home) — `URL.path` strips the `isDirectory: true` slash every time. Still a computed `static var`, so still evaluated per call |
| `captureSessionID` body | `SurfaceRef.zmxSessionPrefix == "graphcode-"`; same emitted string, `singleQuoted` applied once in the same positions |
| Callers of `json(...)` / `captureSessionID(...)` | Exactly three repo-wide: `PresenceHooks.write` (local), `remoteWriteFragment` (remote), tests. No target builds hook JSON independently |
| `hooksFile` side-effecting write | Still fires on the local branch. On `main` the remote case returned before reaching it, so no call is lost or gained on either side. `reportingPath` and `remoteHooksSuffix` collapse to `main`'s values when `remote == nil` |
| Placeholder carve-out | Whole-string `==` on a sentinel, reachable only from the two remote-only callers |
| Sweep filter | `RemoteProjectLocation.parse` requires `ssh://`; `graphcode://global` and every folder path return nil. The sweeper is only *started* from a remote-path store creation |
| Tests spawning real ssh/zmx | Zero. The one `ProjectRegistry(...)` in tests opens only local temp dirs, so the sweep is never reached |

### Fixed in `20369b5`

**1. 🔴 A stale ID trapped the loop permanently.** The branch keyed on the id file *existing*, never on whether its transcript still did — and nothing ever deleted the remote file (`kill` removes only the local one; the remote path returns before `SessionIDStore.remove`). An expired transcript meant `claude --resume` exits, the session dies, and every sweep tick retried the same dead ID with the fresh branch unreachable forever.

The review's suggested `|| { rm; fresh; }` doesn't work here: `zmx run -d` returns 0 the moment the detached session exists, long before the agent inside it fails. So the ID is now **consumed before the attempt** — a resume that lands has its `SessionStart` hook write it straight back, and one that doesn't costs a single wasted launch before the next tick starts fresh.

**2. 🟠 Overlapping ensures could double-launch one session.** The single-shell `get || run` closes the window inside one dial, not between two. A hung ssh dial is bounded well past 60s by `ConnectTimeout` + `ServerAlive` + `runRemoteRetrying`'s backoff, so it is still in flight at the next tick; two dials whose checks both miss both run, and the second types the entire launch command into the agent the first one started. New `RemoteEnsureGate` — one ensure per node at a time.

**3. 🟡 Per-tick cost was much higher than stated.** The hooks write and file delivery ran *before* the check, so every sweep tick re-sent ~20 KB of base64'd shim through a `python3` (plus a second for Copilot's trust seed) for sessions that were already running. Both now sit behind the check. Create-only also matches what `PresenceHooks.write` already documents — an upgraded graphcode reaches a loop at its next session, not mid-run.

**4. 🟢 `remoteSweeper` was never cancelled.** Now cancelled in `deinit`, matching `GraphStore`'s goal pollers.

**5. 🟢 Reader and writer restated the sessions path separately.** `remoteSessionIDExpression` now composes from `remoteSessionsExpression`; divergence would have made every remote resume fall *silently* through to the opening prompt.

Two comments the review showed were wrong were also corrected: the sweep is not sequential (`ensureSession` is fire-and-forget by contract), and its cost was understated.

### Verification

- Full suite: **726 passing**, exit 0 (up from 723 — 3 new cases, 2 existing ordering assertions re-anchored from `zmx get` to `zmx run` now that the preamble sits behind the check).
- `zsh -n` on the restructured script and its quoted wrapper — both parse.
- All four branches exercised against a stub `zmx`: id present → `--resume sess-abc-123` and the file consumed; next tick → fresh prompt (trap closed); hostile payload `x; touch /tmp/PWNED` → arrives as one inert argv word, nothing executed; live session → nothing runs, id file kept.
- `swiftlint` / `swift format lint`: no new violations.

### Known, not fixed

A resumed session comes back **alive but inert** — `resumeArguments` passes `prompt: nil`, so the transcript is restored and nothing is typed in. This is symmetric with the pre-existing local resume path, so it is not a regression, but "resumes" here means the context is back, not that the loop re-fires on its own.

scgopi:
## Second review pass — 2 defects, both fixed (`299bd8f`)

The re-review cleared consume-on-use **empirically** and found two defects the previous commit introduced. Both are fixed.

### ✅ Consume-on-use verified by running Claude Code, not by reasoning

I flagged this as the risky assumption in `20369b5`: deleting the ID before attempting `--resume` only works if `SessionStart` fires on resume and re-exports `$CLAUDE_CODE_SESSION_ID`. If it didn't, every remote loop would resume exactly once and then silently revert to fresh launches forever — worse than the trap it replaced.

Tested against `claude` 2.1.226 with a settings file matching the exact shape `PresenceHooks.json` emits:

| Run | Hook fired? | ID written | Exit |
|---|---|---|---|
| Fresh | ✅ | `90f3ec12-…c8`, equals the reported `session_id` | 0 |
| `--resume 90f3ec12-…c8` | ✅ **fires again** | **same ID**, itself valid for the next resume | 0 |
| `--resume 00000000-dead-…` | ❌ never fires | nothing | **1**, `No conversation found with session ID:` |

The round-trip closes and chained resumes work indefinitely. Run 3 also independently confirms the defect the fix targets — a stale ID exits immediately and fires no hook, so nothing would ever have rewritten the file. Docs corroborate (`SessionStart` runs again on resume with `source: "resume"`; an omitted matcher activates on every occurrence), and `cleanupPeriodDays` defaulting to 30 days makes the stale-transcript premise documented rather than hypothetical.

⚠️ Recorded, not actioned: `CLAUDE_CODE_SESSION_ID` is **undocumented** — only `session_id` on the hook's stdin JSON is guaranteed. The whole resume mechanism, local and remote, rests on an unversioned env var.

### 🔴 Fixed: the gate could silence a node permanently

`RemoteEnsureGate` was strictly worse than what it replaced. `runRemoteRetrying` → `PTYProcessSession.waitCollectingOutput` ends only when the child's `terminationHandler` finishes the stream, and there is **no timeout anywhere in that chain**. With `ControlMaster=auto`, a *wedged* master — laptop slept mid-session, Codespace stopped holding the socket — blocks a new client on the unix socket, where neither `ConnectTimeout` nor `ServerAliveInterval` applies. So `end()` was unreachable and the entry never cleared: every later ensure returned immediately and that loop was **silently dead until the daemon restarted**.

Before the gate, the same wedge leaked one task per tick but the node still got a fresh dial every 60s. The gate turned a self-healing waste into a permanent stop — the same class of bug this whole branch exists to eliminate.

Now a **5-minute lease** rather than a flag: far longer than any healthy dial (3 × `ConnectTimeout=10` + 1s/2s backoff), short enough that a wedge costs minutes not the process lifetime. Re-dialling after expiry is safe — a dial wedged before reaching the host ran nothing, and one wedged after `zmx run` leaves a session the next `zmx get` finds.

### 🟠 Fixed: a graphcode upgrade couldn't reach a host with live loops

Moving delivery behind the existence check had a casualty I hadn't listed. The trust seed and hooks file are genuinely create-only — a folder-trust dialog is answered per session start, and Claude Code reads `--settings` once at startup. But the **CLI shim is re-executed throughout a session's life** and speaks `FramedMessageIO` framing and `DaemonProtocol` JSON to this daemon. A host left on an older copy breaks `graphcode node send` / `memo` / `resolve` for every loop already running there — and an unattended loop has no human to open it and heal the delivery.

Delivery is now gated on a **content stamp** instead of session existence: it runs when the session is missing (the create branch launches an argv naming briefing/wake/prompt files that ride in the same fragment) **or** when the stamp differs. A healthy tick costs one extra `zmx get` and a `cat` on the same host, so defect #3 from the first pass stays fixed. FNV-1a rather than CryptoKit — the question is only "same bytes", and Swift's `hashValue` is seeded per process, so it would report a change on every daemon restart.

### Also: two weakly-anchored assertions strengthened

- The hooks test matched `.graphcode/hooks`, which **also** occurs in the launch argv as `--settings "$HOME/.graphcode/hooks/claude-code.json"` — it would have passed with `remoteWriteFragment` deleted entirely. Now anchored on `mkdir -p "$HOME/.graphcode/hooks"`.
- Ordering assertions verified textual position only; `check || { }; seed; run` would have satisfied them. They now pin the fragment *inside* the `|| { … }` group.

### Verification

- Full suite: **730 passing**, exit 0.
- `zsh -n` on the restructured script and its quoted wrapper.
- Stub-`zmx` runs covering every branch: dead + unstamped → deliver, seed, hooks, launch; **live + current stamp → nothing at all**; **live + stale stamp → deliver only, no launch, stamp healed**; resume consumes the ID → next tick falls through to fresh; hostile ID payload inert.
- `swiftlint` / `swift format lint`: no new violations.

### Cleared with no finding

Gate early-return/throw/cancellation paths (none exist — `startRemote` has no `throws`, and every `await` is cancellation-safe); non-sweep callers (`createNode` and `spawnInstance` always carry fresh UUIDs; composites aren't swept; load-time ensure precedes the first tick); reopening a loop in the app (the app's remote path is `GhosttyTerminalView.remoteCommand` + `zmx attach`, never `ZmxSessionLauncher.start`, so no human action can be blocked); staged messages to sleeping remote loops (a sleeping loop takes the create branch, where delivery still runs); the local path across the combined `main...HEAD`.

scgopi:
## Third review pass — blocker reproduced and fixed (`52a929e`)

### 🔴 The stamp-in-manifest fix didn't close the hole — reproduced, not argued

`installerScript`'s comprehension iterates `sorted(m.items())` and is **not transactional**, and `.` (0x2E) sorts before `g` (0x67):

```
~/.graphcode/bin/.shim-stamp     <-- written FIRST
~/.graphcode/bin/graphcode
```

So "one python process writes shim and stamp or neither" was not what the code did. On a Codespace near its disk quota the 16-byte stamp write succeeds and the 14.6 KB shim write hits `ENOSPC`; the non-zero exit is swallowed by `|| true`, the matching stamp then skips every later delivery, and **the host is permanently un-upgradeable**. Same class as the `printf` it replaced, just narrower — and the trigger is realistic precisely *because* the small write succeeds and the big one doesn't.

The stamp is now a **receipt**: `installerScript` takes it separately and writes it in a statement *after* the comprehension, which a raise anywhere above skips outright.

Verified by running the real emitted python program against a manifest whose shim write is forced to fail:

| Case | Shim on disk | Receipt |
|---|---|---|
| Shim write fails | `OLD SHIM v1` | **`<absent>`** → next dial retries |
| Shim write succeeds | `NEW SHIM v2` | `stampV2` |
| No receipt argv | `NEW SHIM v2` | `<absent>` (guard holds) |

### 🟡 The lease could be released by someone else

`end(_:)` removed unconditionally, with no way to know the lease was its own. Dial A wedges → its lease expires → dial B takes a fresh one → A finally returns and deletes **B's** lease, leaving B in flight with the gate open. The next tick starts a third dial; with the host just back, both miss `zmx get` and both `zmx run` — the exact leak the gate exists to prevent, in the wedge-recovery case it was written for. `begin` now returns the lease as a token; `end` releases only on match.

### 🟢 Renamed `deliveryStampPath` → `shimStampPath` (`.delivery-stamp` → `.shim-stamp`)

The stamp hashes the shim alone while gating a manifest that also carries the briefing, wake digest and prompt. That scope is *correct* — the wake digest changes on nearly every ensure, so hashing the manifest would fire delivery every tick and defeat the point — but the old name invited the next person to assume otherwise. The doc comment now states what it covers and that everything else stays correct via the `! check` disjunct.

**Also:** the previous commit's test didn't test its own name. `aFailedDeliveryLeavesNoStampBehind` asserted the stamp was in the manifest and that no `printf` existed — neither of which can catch write *ordering*. It now pins the receipt outside the manifest and after the comprehension.

### Cleared with no finding (verified by execution)

- **Tilde expansion** — `~` inside `$( )` inside `" "` expands in zsh, bash, dash and `/bin/sh`, directly and through a nested single-quoted `-c` wrapper, in both directions (matching stamp → skip, stale → deliver). Agrees with `os.path.expanduser` both when `HOME` points somewhere other than the passwd entry and when `HOME` is unset.
- **Lease duration** — worst-case *healthy* dial computes to ~78s (3 × (`ConnectTimeout` 10 + `ServerAlive` 5×3) + 1s + 2s backoff), ~100–110s with a fat remote `zsh -l -i` startup. 300s is ~3× headroom while capping a wedge at five missed ticks. Expiry alone cannot reintroduce the double-`zmx run`: the second dial re-runs `zmx get` on the host, so both would have to run within the millisecond gap between one dial's own check and run.
- **Fresh launch with a matching stamp** — the `! check` disjunct fires, so briefing, wake digest and PROMPT.md are re-delivered as the argv about to run requires. All four gate states exercised against a stub.
- **`GhosttyTerminalView.remoteCommand`** — untouched, still delivers unconditionally; the extra argv adds ~24 bytes and now also stamps the host on a human attach, letting a later ensure correctly skip a redundant delivery.
- **Local path** — `RemoteEnsureGate`, `deliveryFragment`, `cliShimStamp` and `shimStampPath` are remote-only by grep; nothing touches `localSessionsExpression`, `json(forBackend:zmxPath:)`, `resumeArguments`'s `remote == nil` branch, `SessionIDStore`, or `atomicCheckOrRun`. Byte-identical verdict stands for the full `main...HEAD`.
- **FNV-1a** — correct offset basis and prime, wrapping multiply, process-independent.

### Verification

**732 passing**, exit 0. `swiftlint` / `swift format lint`: no new violations (one pre-existing serious violation in `ProjectFeature.swift`, unrelated).

scgopi:
## Correction to my third-pass comment above

The review's crossed follow-up corrected a factual claim I made, and it's worth stating plainly rather than leaving buried.

**I named the wrong trigger.** My comment above (and the code comment in `installerScript`) said a Codespace near its disk quota was what strands a receipt beside a stale shim — the 12-byte stamp write succeeding where the 14.6 KB shim write hits `ENOSPC`. Measured against a forced-full volume, that is close to unreachable: **both** writes fail and nothing is stranded. It would need the freak case of enough free blocks for 12 bytes but not 14.6 KB.

The **deterministic** trigger is permissions — a shim that can't be overwritten under a directory that can. An ordinary Codespace shape: `graphcode` owned by another uid, a devcontainer feature's own copy, `chattr +i`. That is also exactly what my reproduction used (`chmod 400` on the shim), so the fix was always verified against the real trigger — I just described it wrongly.

Corrected in `1d2bff4` (comment-only, 732 still passing). Left as-is, it would send the next person looking at disk space when the issue is file ownership.

**Also recorded:** ordering alone is sufficient, with no `with`/`flush` bookkeeping — a payload this size raises at `.write()` rather than at an implicit close, leaving no file behind, so the trailing receipt statement is genuinely unreachable on failure.

**Not taken:** the reviewer's alternative of moving the marker *inside* the shim as a trailing `# graphcode-shim-stamp: <fnv>` line, with the dial doing `grep -q` instead of `cat`. It's a good idea — the stamp then cannot diverge from the shim because it *is* the shim, and it dissolves the naming concern entirely. Deferred as a follow-up rather than a sixth round on this branch.

PR #76, opened 2026-08-09
docs: put cross-vendor orchestration in the pitch, not just the code

scgopi:
## What

Three surfaces named a single vendor. Every one of them is above the fold:

| Surface | Before | After |
|---|---|---|
| `docs/_layouts/default.html` — hero tagline on **every page** of graphcode.app | "Graphs of live, steerable **Claude Code** sessions on macOS." | "Graphs of live, steerable **coding-agent** sessions on macOS — Claude Code, Codex and Copilot CLI loops on one graph." |
| `README.md` — opening block | no mention of other backends | one sentence: an edge can hand work from a Claude Code loop to a Codex one |
| `docs/index.md` — front-matter `description` (the search/link-card line) | "…for coding agents on macOS" | names the three backends |

## Why

Running ten agents in parallel is the claim every competitor in this category already
makes. A graph whose **edges cross vendors** is the one neither Anthropic nor GitHub can
answer with a point release — and it was in the code and absent from the pitch.

`graphcode.app` is currently the **#2 referrer** to this repo (187 views / 26 uniques,
14d), behind only `github.com`. That hero line is the highest-traffic sentence the
project owns.

## The claim is bounded by what the code actually does

From `GraphcodeKit/Sources/Domain/BackendCapabilities.swift`:

- All three rows are spiked (`isSpiked` is `true` for `claudeCode`, `copilotCLI`, `codex`).
- All three set `supportsMidSessionInput: true`, which is the capability gating the `to`
  side of a `.message` edge; `.handoff` and `.spawn` are strictly more permissive because
  they start a session rather than interrupt one.

Nothing here claims parity between backends — "Claude Code is the most complete backend"
stays exactly as written further down `index.md`.

## Not done here, flagged for a decision

The GitHub repo description still reads *"…and work that continues without you."* That
sentence describes something the CLI vendors now ship themselves (mobile push on
needs-input, Remote Control). It's a settings-page edit rather than a file in this diff —
worth changing to lead with the graph, but that's your call, not a silent push.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://graphcode.app/">
-    <img src="docs/assets/banner.png" alt="GraphCode — graphs of live, steerable Claude Code sessions on macOS" width="100%">
+    <img src="docs/assets/banner.png" alt="GraphCode — graphs of live, steerable coding-agent sessions on macOS" width="100%">
   </a>
 </p>
 
@@ -22,6 +22,10 @@ connected, unattended, and still yours to attach to and correct mid-run. Each no
 graph is a unit of work running inside a real CLI coding-agent session; each edge is a
 hand-off, message, or spawn between them. The sessions are real terminals you can attach
 to, watch, and steer — not headless jobs that report back when they're done.
+
+The nodes don't have to be the same agent. Claude Code, Codex and Copilot CLI all host
+loops and all take input mid-session, so an edge can hand work from a Claude Code loop to
+a Codex one on the same graph — one graph, spanning vendors.
 
 **[Graph Engineering, simplified →](https://graphcode.app/)** — the full
 article: the mental model, then the machinery underneath it.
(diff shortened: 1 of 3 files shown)


scgopi:
Added a second commit rather than opening a separate PR, so this stays one review.

`#78` (merged today) gave Copilot CLI and Codex loops live activity reporting, each read from the log that CLI already writes. The site's **What runs today** paragraph still described the world before it — and while checking it I found a claim that has never been true:

> Copilot CLI and Codex loops launch, run, **fan out on their own provider**, message, and keep memory like Claude ones.

`canHost(.composite)` requires `supportsSubAgents`, which is `false` for both. Fan-out is the one thing they can't do.

The replacement states the real matrix instead: composites and hook-based presence are Claude Code only; Copilot adds time-based loops over Codex (`supportsInSessionRecurrence`); all three now report the tool call they're inside; and remote Copilot/Codex activity stays `nil` because the log is on the far host — the limitation #78 shipped with, now written down where users read it.

PR #102, opened 2026-08-13
Summary rails: a loop's narration, read off its own transcript

scgopi:
A new section at the top of the workspace rail — what the loop is doing now in one sentence, the two beats before it receding up a flowing spine, a pass line where the metric moved, and one row per finished pass down to a count. Built from `design review/summary_rail_handoff`.

## The producer calls no model

All three backends narrate before they act, and all three write that narration to disk. A beat is a tail read of a file graphcode is already reading for `activity` — zero tokens, zero subprocesses.

| Backend | File | Narration → beat text | Tools → kind + evidence |
| --- | --- | --- | --- |
| Claude Code | `~/.claude/projects/<p>/<session-id>.jsonl`, found by the id already banked for `--resume` | `assistant` → `text` blocks | `tool_use` |
| Codex | `~/.codex/sessions/**/rollout-*.jsonl` | `agent_reasoning`'s `**bolded header**` | `function_call` |
| Copilot | `~/.copilot/session-state/<id>/events.jsonl` | `assistant.message.content` | `tool.execution_start` |

Tool phrases come through each backend's existing `phrase` mapping, so a beat's evidence names exactly what the card's live line would have. A new narration closes the open beat — that is what makes twenty greps in a row one beat, and a pass is a user turn (a human typing, or `/loop` waking the session).

Bounded by construction: three beats, three pass summaries, a count for the rest. A six-hour loop and a six-minute one are the same height and the same bytes in the graph file.

## Two settings, both off

- **Summarise what loops are doing (experimental)** — the producer. Off, nothing is read, no node carries a summary, and the card's live line falls back to what the loop was handed, exactly as before.
- **Let a model write the summaries** — the optional second pass: one `claude -p` / `copilot -p` / `codex exec` at the fast tier, per *changed* beat of a *working* loop, over one sentence and one line of evidence — never the transcript. Its own process, so its tokens never land on the loop's own budget line; every failure path leaves the agent's own wording standing. Flags checked against each CLI's `--help`, not assumed.

## What the section does

Four states — starting, working, asking, resolved. `asking` is read off `displayState`, never off the summary, so attention keeps one source of truth. Folding keeps one truncated line; ⌥G hides the rail and leaves a 26pt gutter that still turns amber when the loop wants a human. `SINCE YOU LOOKED` is per window, set on the way out, so coming back is what shows it.

The spine's upward flow is the one `repeatForever` animation in the app, guarded the way `StateIndicator`'s removed pulse was not: one per window, only while the session is busy and the section is open.

## Checks

- 838 tests pass (34 new across three suites), `swiftlint lint` and `swift format lint --strict` clean.
- Store bounding asserted across a 500-beat run; pass rollup asserted to happen once and never be rewritten; card fallback asserted with the producer off; old graph files and old settings files asserted to still decode.

### DIFF
--- a/GraphcodeKit/Sources/Domain/GraphcodeSettings.swift
+++ b/GraphcodeKit/Sources/Domain/GraphcodeSettings.swift
@@ -246,6 +246,36 @@ public struct GraphcodeSettings: Codable, Equatable, Sendable {
   /// is a trade worth offering and not worth imposing.
   public var showsActivityStrip: Bool
 
+  /// Whether graphcode narrates what loops are doing — the summary rail's *producer*.
+  ///
+  /// **Off by default, and experimental.** Off, nothing reads a session's transcript, no
+  /// node carries a `summary`, the rail shows no section and the card's live line falls
+  /// back to what the loop was handed, exactly as it did before this existed. On, each
+  /// working loop's transcript tail is read once per poll and folded into
+  /// `LoopNode.summary`.
+  ///
+  /// Deliberately distinct from hiding the rail. Hiding is a view, and this is the
+  /// reading: the design's own division, kept because the two would otherwise be one
+  /// switch that quietly does two things. It is a setting at all — rather than always-on —
+  /// because a beat is a claim about what an agent is *trying* to do, and a wrong one is
+  /// worse than the scrollback it replaces. It stays behind a switch until it has been
+  /// watched on real loops.
+  public var summarisesLoops: Bool
+
+  /// Whether a small model may rewrite the current beat, on top of the free reading.
+  ///
+  /// **Off by default, and meaningless with `summarisesLoops` off.** The rail works
+  /// without this: the beat is the agent's own sentence, taken off its transcript for
+  /// nothing. What a model buys is the last edit — the design's plain past tense, ten
+  /// words, tools turned back into intent — and it is the one part of the feature that
+  /// costs money, so it is the one part behind its own switch.
+  ///
+  /// One `claude -p` / `copilot -p` / `codex exec` per *changed* beat of a *working* loop,
+  /// at the fast tier, over one sentence and one line of evidence. It is its own process,
+  /// so its tokens never land on the loop's own budget line, and every failure path leaves
+  /// the derived beat exactly as it was — see `SummaryModelWriter`.
+  public var summaryUsesModel: Bool
+
   // There is deliberately no window-opacity setting here any more. graphcode used to own
   // one and apply it as `NSWindow.alphaValue`, which fades the whole window — terminal
   // text included — rather than only the background behind it. Ghostty already has
@@ -261,6 +291,8 @@ public struct GraphcodeSettings: Codable, Equatable, Sendable {
     briefsSessionsAboutTheGraph: Bool = true,
     autoSelectsModel: Bool = false,
     showsActivityStrip: Bool = false,
+    summarisesLoops: Bool = false,
+    summaryUsesModel: Bool = false,
     worktreePolicies: [String: WorktreeHygienePolicy] = [:]
   ) {
     self.defaultBackend = defaultBackend.isSpiked ? defaultBackend : .claudeCode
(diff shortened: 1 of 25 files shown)


scgopi:
## A second pass, driven off real transcripts rather than fixtures

Three commits since the review above. Everything here was found by running this machine's own Claude Code, Copilot and Codex logs through the readers — a 108 MB transcript is a different animal from a fixture, and that is where all of it came from.

### `2366e71` — six faults the rail showed on real sessions

| Fault | What it looked like |
| --- | --- |
| Pass numbers counted inside the tail read | A session on its 41st turn said `PASS 10`, and the number walked **backwards** as the file grew past the turns that opened it. The store now owns an absolute `currentPass`, advanced once per turn it has never seen; readings are shifted onto it. |
| Beat ids carried the pass | Every renumbering re-animated the whole rail and lost the `SINCE YOU LOOKED` watermark. Ids are the beat's own instant now. |
| The hairline sat one row too high | `unseen` counts the current beat, which is the block under the rows, not one of them — so a returning human saw the last beat they *had* read marked new. |
| Switching the producer off froze a beat on every card | The reader answers an empty reading rather than `nil` when the switch is off, and the store reads that as "this node carries no summary". Off clears; it used to freeze. |
| A link's whole address reached the rail | `[@name](https://…)` took the line; cuts landed mid-word (`Threads/Instagr…`). Labels survive, addresses don't, cuts land between words. |
| The hidden-rail gutter went amber `ASKS YOU` | At a session sitting quietly at its prompt — it was the last surface still reading `displayState`. |

Plus: summary readings are taken concurrently. They are the one reading with an optional subprocess behind them, and sequentially that was one timeout *per loop* on the tick every state dot rides on. The model timeout is now under the poll interval.

### `3cc0203` — the rail speaks the app's ten languages

It was the one corner still answering in English. Three strings needed to *reach* the catalog rather than just have an entry in it: `Text` localises a `LocalizedStringKey` and prints a `String`. 31 keys × 10 languages, checked against the compiler's own extraction — every key the build emits for these files has an entry, bar `p%lld`, which is a 26pt mono column and stays `p7` everywhere. A beat is never translated: the agent narrates in whatever language it is being talked to in.

### `547e443` — the number belongs to the pass it is printed under

Deltas were keyed by position, sample *n* to pass *n*. True of a `/loop` on a timer, false the moment a human types twice in one pass. Matched on when the sample was taken now; a pass with no sample inside it says nothing.

847 tests pass (nine new), `swift format lint --strict` clean.

scgopi:
## Performance and regression pass

### The switch now takes effect in the window, not a poll later

Reported from a real session: with `Summarise what loops are doing` **off**, rails were still on screen. The daemon-side fix earlier in this PR clears the node, but that lands on the next poll — and never at all for a graph file written by a build that cleared nothing. So the window now asks the same question, off its own live copy of the settings: the section, the 26pt gutter it collapses to, `LoopWorkspaceRail.hasContent`, and the card's live line all go the moment the toggle flips, folded or expanded. Two answers to one question, agreeing, fifteen seconds apart at worst.

### Four costs, measured on a 113 MB transcript

| Cost | Before | After |
| --- | --- | --- |
| Tail read + parse (per changed transcript, per poll) | 11.1 ms | **6.8 ms** |
| Resolved loops | asked every poll — a directory walk per backend, and Codex's is over every rollout on the machine | asked only while they still carry a summary to clear |
| A timed-out model rewrite | left `claude -p` running on its own PTY, one per poll per loop | `terminate()` on the way out |
| The section's presentation | rebuilt twice per frame, on every clock tick | once |

The read was decoding 512 KB into a `String` to split it, then **re-encoding** every kept line for `JSONSerialization` — most of it spent on one 400 KB tool result that is dropped as a fragment two lines later. It splits on the newline byte now and hands the slices straight through.

### Two leaks

- A deleted loop kept a modification date in the freshness cache for the life of the daemon.
- `NSCursor.push`/`pop` on the rail's resize handle only balanced if the pointer left the way it came. Hiding the rail with ⌥G mid-hover left the resize arrows following the pointer around the whole app.

848 tests pass, `swift format lint --strict` clean.

PR #156, opened 2026-08-23
Workspaces: separate graphs, daemon and loops per window

scgopi:
## What this solves

One sidebar of loops stops being monitorable somewhere past a few dozen, and unrelated lines of work have no reason to share a graph. This adds **workspaces**: a support directory of its own — `~/.graphcode-<name>`, beside the default `~/.graphcode` — with its own projects, loops, terminal layouts and `graphcoded`.

**File ▸ Workspace ▸ New Workspace…** names one and opens it. It is a second app instance, so it gets its own Dock tile and can live on a second screen.

## Why it needed more than the environment variable

`GRAPHCODE_SUPPORT_DIR` has always been able to point a second graphcode elsewhere. Two things made it unusable as a product feature, both fixed here:

- **One hardcoded launch agent label.** A second app rewrote `dev.graphcode.graphcoded` and took the first workspace's daemon with it. Labels are now per workspace (`dev.graphcode.graphcoded.<slug>`).
- **launchd starts an agent with a minimal environment**, so the daemon it spawned computed `~/.graphcode` whatever the app was pointed at. A named workspace's agent now carries its directory in `EnvironmentVariables`.

## The default workspace is untouched

Same path, same label, and a plist with **no new keys** — an extra key there would make `launchAgentIsCurrent` report every existing install as stale and bounce a daemon that was running perfectly well. `theDefaultWorkspacesAgentIsUnchangedAndANamedOnesCarriesItsDirectory` pins both halves.

A side effect worth naming: a developer running `GRAPHCODE_SUPPORT_DIR=~/.graphcode.dev` shared the installed app's label until now, so launching that build rewrote the real agent and pointed launchd at a DerivedData daemon. Every non-default directory now gets a label of its own.

## Also here

- **A pid file per workspace** — opening one already open raises its window instead of starting a second app over one set of graphs and `zmx` session names.
- **Socket paths validated at naming time** against `sun_path`'s 104 bytes, with the numbers in the message, rather than failing to bind later and looking like a dead daemon.
- **A switcher at the foot of the sidebar** — the only thing on screen that says which workspace a window is; two instances share an icon and have no titlebar. Shown only once there is more than one workspace.
- Workspaces are found by scanning for `~/.graphcode-*` rather than kept in a registry file, which would have to live in one workspace and be written by all the others.

## Verification

- `xcodebuild … test` — **984 tests, all passing**; 12 of them new.
- `swift format lint --strict` clean over `GraphcodeKit` and `graphcode`.
- Launched a second instance of the built app through the same `NSWorkspace.OpenConfiguration` path the client uses, pointed at a probe directory, and confirmed on the live machine: `GRAPHCODE_SUPPORT_DIR` arrives in the child (with `PATH` intact — the configuration adds to the environment rather than replacing it), `app.pid` lands in the probe directory, no launch agent is written for an unpackaged build, and the installed workspace's agent checksum and running daemon are unchanged.

## Not in this PR

Deleting or renaming a workspace (remove the directory by hand for now), and any sharing between workspaces — a loop in one is invisible in the other, which is the point.

### DIFF
--- a/GraphcodeKit/Sources/DaemonBootstrap.swift
+++ b/GraphcodeKit/Sources/DaemonBootstrap.swift
@@ -27,7 +27,10 @@ public enum DaemonBootstrap {
     case failed(String)
   }
 
-  private static let label = "dev.graphcode.graphcoded"
+  /// One agent per workspace, so that opening a second one installs a daemon of its own
+  /// instead of rewriting the first's. The default workspace keeps the bare label it has
+  /// always had — an existing install must not see its agent renamed.
+  static var label: String { Workspace.current.daemonLabel }
   private static let appBundleIdentifier = "dev.graphcode.app"
   /// `graphcode` here is the CLI, not the app — a different product that happens to share
   /// the name a human types. Shipping it matters: `~/.graphcode/bin` is what the README
@@ -167,11 +170,20 @@ public enum DaemonBootstrap {
 
   /// Built here rather than shipped as a template file: the two paths it needs are already
   /// known at runtime, and a template would be one more thing to keep in sync.
+  ///
+  /// `workspace` decides both the label and whether the daemon is told where to look.
+  /// launchd starts an agent with its own minimal environment, so a named workspace's
+  /// daemon inherits nothing from the app that installed it and would compute
+  /// `~/.graphcode` — serving the default workspace's graphs under the second
+  /// workspace's label. The default workspace passes no `EnvironmentVariables` at all,
+  /// which keeps its plist byte-for-byte what it has always been: anything else and
+  /// `launchAgentIsCurrent` would report every existing install as stale and bounce a
+  /// daemon that was running perfectly well.
   static func launchAgentPlist(
-    daemonPath: String, supportDirectory: String
+    daemonPath: String, supportDirectory: String, workspace: Workspace = .current
   ) -> [String: Any] {
-    [
-      "Label": label,
+    var plist: [String: Any] = [
+      "Label": workspace.daemonLabel,
       "ProgramArguments": [daemonPath],
       "RunAtLoad": true,
       "KeepAlive": true,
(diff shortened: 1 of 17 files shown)


scgopi:
Added in `5a109b4`: **deleting a workspace**, and **keyboard shortcuts**.

### Delete Workspace

By hand this is a trap — the agent is `KeepAlive`, so a daemon left loaded over a removed directory is restarted by launchd and recreates it, and `zmx` outlives every app, so the loops keep running against a graph that no longer exists. The menu item does the four steps in the order that works:

1. Boot the daemon out of launchd and remove that workspace's agent.
2. End every `zmx` session named in the workspace's graphs **and** its terminal layouts — extra tabs and splits are surfaces of their own, and a missed one is an agent running until the machine reboots.
3. Move the directory to the **Trash**, not `rm -rf`. It is the only copy of that work.
4. Refresh the switcher.

Refused for the default workspace, for the one the window is using, and for one another instance has open — each with a reason rather than a disabled row. The confirmation names what goes with it ("It holds 2 projects and 14 loops").

### Shortcuts

| Keys | Does |
|---|---|
| ⌥⌘1 … ⌥⌘9 | Switch to that workspace in menu order (⌥⌘1 is the default one) |
| ⌥⌘N | New Workspace… |

⌘1…9 are the terminal's tabs and ⌘⇧3/4/5 the system's screenshots, which leaves ⌥⌘ the free row. Declared only on the menu bar's copy of the items — the sidebar popup renders the same list without them, since a key equivalent declared twice is one macOS resolves by picking a winner.

### Also

`WorkspaceLock.claim` now declines when a live process already holds the workspace. Claiming unconditionally meant any second process touching it — the test host launching the app, for one — replaced a running window's pid, and once that process exited the workspace read as free while the window still had it. That was reaching `~/.graphcode`, so it counted as breaking the default workspace.

**987 tests passing** (16 workspace tests), `swift format lint --strict` clean.

### Known edge

A `graphcoded` started by hand rather than by launchd is not stopped by the delete — nothing tracks those. The launchd-managed one, which is every install, is.

Issue #167, opened 2026-08-23
graphcoded takes ~29s to exit under launchctl bootout


scgopi:
Found while fixing #166. Deleting a workspace boots its daemon out of launchd, and the app's action log showed `launchctl bootout gui/<uid>/dev.graphcode.graphcoded.personal7` taking ~29 seconds — the signature of launchd escalating past SIGTERM because the process did not exit on it.

The app no longer blocks on this (the tear-down runs off the main thread since #166), but it still means:

- deleting a workspace takes half a minute to actually reach the Trash
- every app-update daemon reload (`DaemonBootstrap.reloadDaemon`) pays the same wait
- a reboot's shutdown sequence may be killing graphcoded rather than letting it exit

Likely cause: `graphcoded` has no SIGTERM handler that breaks its accept loop — see `graphcoded/Sources/main.swift`, whose socket loop runs on a detached queue with nothing listening for termination.

Fix shape: install a signal source for SIGTERM that closes the listening socket and exits cleanly, so `bootout` returns in milliseconds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #200, opened 2026-08-27
graphcoded exits when its binary is replaced; updates reach closed workspaces

### LINKED ISSUE #199: graphcoded never notices its binary was replaced — stale daemons run old code indefinitely
## Problem

Nothing compares the running `graphcoded` image to the binary installed at its path. `DaemonBootstrap.installIfNeeded()` boots the daemon out only when it runs — at app launch, for the current workspace, and only when the helper stamp differs. Every path that misses that call leaves a stale daemon running old code with launchd's `KeepAlive` faithfully keeping it alive:

- **In-app updates without an effective relaunch.** The updater's relaunch was broken until 0.1.50-beta3 (`open -n` from a detached shell never ran), so every "Install Update" advanced the on-disk version while the old app *and old daemon* kept running. Users report the version Finder shows — not the version the daemon runs.
- **Secondary workspaces.** Each workspace has its own daemon, upgraded only when that workspace's window is opened. A workspace left closed keeps its old daemon forever.
- **Declined relaunch prompts / DMG-over-app installs** while the app keeps running.

This is how the pre-0.1.44 PTY leak (#151's fix, `PTYProcessSession`) kept leaking on machines whose on-disk version said 0.1.51: the leaky *process* survived every upgrade. Verified during the #196 investigation — leaked fds die with the daemon process, so a stale daemon is the only way old leaks persist.

## Fix shape

`graphcoded` records its own executable's identity (device + inode + size — the install swap always lands a fresh inode) at startup and re-stats it on a timer. When the file at its launch path is a different file, it exits cleanly the same way the SIGTERM path does (unlink socket, `exit(0)`), and launchd's `KeepAlive` respawns — exec'ing the new binary at the same path. A path that can't be stat'd (relative argv[0] from a hand launch, transient window mid-swap) skips the tick rather than exiting.

This closes every stale-daemon path at once, including workspaces whose windows are never opened, with one `stat` per minute.

scgopi:
Fixes #199. Companion to #198 — this closes the *other* mechanism behind the 0.1.51 PTY-leak reports: stale daemon processes still running pre-0.1.44 leaky code under a new on-disk version.

## Daemon half: staleness self-check

`graphcoded` records its own executable's identity (`ExecutableIdentity`: device + inode + size) at launch and re-stats it once a minute. `DaemonBootstrap`'s install always stages a copy and renames it into place, so an update is always a fresh inode. On a mismatch the daemon exits cleanly exactly like the SIGTERM path (socket unlinked, `exit(0)` from the main queue), and launchd's `KeepAlive` respawns the path — now the new binary.

Safety posture: an identity that can't be read at launch (relative argv[0] from a hand launch) disables the check; an unreadable tick (the rename window mid-swap) is skipped, never treated as "changed". One `stat` per minute.

## App half: closed workspaces get the update too

`installIfNeeded` only ever served the current workspace, and each workspace's daemon respawns from its own `bin` — so a workspace whose window was never opened after an update kept an old daemon (and old helpers) forever. The bootstrap now carries the bundled helpers to every **closed** sibling workspace: installs into its bin, refreshes its launch agent, reloads its daemon, writes its stamp.

- A workspace held by a live app instance (`WorkspaceLock.holder`) is skipped — installing under a running window creates the new-daemon/old-window pairing `changedBundleStamp` exists to prevent; that instance updates itself at its own relaunch.
- A workspace directory with no launch agent has no daemon and is left alone.
- Runs off the main thread: a sibling daemon predating the #167 SIGTERM fix can take launchd ~29s to kill, and app launch must not block on it.
- The sibling reload is what makes the transition work: a sibling daemon older than the staleness timer would never notice its binary changed.

## Verification

- 4 new tests (identity stability, staged-rename detection, unreadable-path nil, install-to-named-destination); full suite 1205 tests in 129 suites green
- `swiftlint` and `swift format lint --strict` clean

https://claude.ai/code/session_01YEP79qiCW9QP6LNZ1F9Wzp

### DIFF
--- a/GraphcodeKit/Sources/DaemonBootstrap.swift
+++ b/GraphcodeKit/Sources/DaemonBootstrap.swift
@@ -133,15 +133,9 @@ public enum DaemonBootstrap {
       // Written somewhere a human (or a bug report) can find: the app has no console,
       // and a bootstrap that failed silently is how a machine ends up looking
       // "corrupted" with nothing to say why.
-      let report = "\(Date().ISO8601Format())  helper install failed: \(error)\n"
-      let log = SupportDirectory.url.appendingPathComponent("bootstrap.err.log")
-      if let handle = try? FileHandle(forWritingTo: log) {
-        _ = try? handle.seekToEnd()
-        try? handle.write(contentsOf: Data(report.utf8))
-        try? handle.close()
-      } else {
-        try? report.write(to: log, atomically: true, encoding: .utf8)
-      }
+      appendReport(
+        "\(Date().ISO8601Format())  helper install failed: \(error)\n",
+        to: SupportDirectory.url.appendingPathComponent("bootstrap.err.log"))
       return .failed("\(error)")
     }
   }
@@ -162,6 +156,11 @@ public enum DaemonBootstrap {
     let expected = stamp(forHelpersIn: bundled)
     guard !expected.isEmpty else { return }
     let current = Workspace.current
+    // An instance pointed somewhere the workspace scan would never find — a
+    // `GRAPHCODE_SUPPORT_DIR=/tmp/…` test daemon, a `~/.graphcode.dev` — is isolated by
+    // intent. Before this feature it touched nothing outside its own directory, and a
+    // refresh from it would break exactly that isolation.
+    guard isStandardWorkspaceDirectory(current.url) else { return }
     for workspace in Workspace.all() where workspace.id != current.id {
       guard WorkspaceLock.holder(of: workspace) == nil else { continue }
       let agentURL = launchAgentURL(forLabel: workspace.daemonLabel)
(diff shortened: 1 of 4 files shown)


scgopi:
Independent adversarial review found 2 major + 3 minor issues; all addressed in the follow-up commit:

- **Major — direction-blind refresh**: any packaged copy (an old DMG in ~/Downloads) could rewrite every closed workspace backward and bounce its daemon on alternating launches. Now newest-wins on the stamp's build modification times (`stampRegresses`), so a stale copy refreshes nothing.
- **Major — isolation break**: an instance launched with a `GRAPHCODE_SUPPORT_DIR` override (or `~/.graphcode.dev`) would have rewritten the user's real workspaces. `isStandardWorkspaceDirectory` now gates the refresh to instances running in `~/.graphcode` or a `~/.graphcode-*` sibling.
- **Major — staging collision**: post-update, every relaunched open workspace refreshes the same closed siblings concurrently, and the shared `<name>.new` staging path let one refresher delete the file another was about to rename (stamp written, daemon missing). Staging is now per-pid with a defer cleanup.
- **Minor**: sibling error log appends instead of truncating (shared `appendReport`); the `ExecutableIdentity` design comment's rename/inode justification was corrected and the benign one-restart-after-remount device-number caveat documented.

Verified-safe by the same review (no change needed): timer exit context matches the #167-safe SIGTERM path; timer retention; relative argv[0] disables the check; `.notPackaged` never reaches the refresh; removed daemons are not resurrected; default-workspace plist parity holds.

2 new tests pin the guards (standard-directory gate, downgrade refusal). Suite: 1207 tests in 130 suites green; format/lint clean.

Issue #217, opened 2026-08-30
UX catalogue: things that confused a first-time user wiring a two-loop graph from the CLI (0.1.56)


Yuncun:
<img width="2000" height="1512" alt="Image" src="https://github.com/user-attachments/assets/b0877f47-166c-449a-af97-3895283951e9" />
<img width="2000" height="1241" alt="Image" src="https://github.com/user-attachments/assets/4529b87a-0592-4861-92f9-9406787d2000" />

Yuncun:
Two more from the first evening of running it:

**13. `--skip-unchanged` can deadlock a loop that is the only thing changing the tree.** Build loop with `--skip-unchanged`: after its turn ended (commit pushed, predicate still failing), the daemon never woke it again for ~60 min — HEAD and the dirty list were unchanged *because the loop itself is what changes them*. The option only makes sense when the predicate watches something external. Suggest a warning at create time when a goal loop has both `--skip-unchanged` and a predicate that reads the repo, or default it off.

**14. `--budget` burns in minutes on a Claude Code session.** `--budget 3000000` was hit after 4 screens / ~5 minutes on Opus, and the node went to **Stalled** (not "budget exhausted"). If the number counts cache-read input tokens per request, 3M is a handful of turns; the help text should say what's counted, and the status should say *why* it stalled.

Also: on stop, both sessions had a suggestion pre-typed into the prompt box (`pull the latest docs commits from main`, `bump the budget and keep going`) that nobody sent — is that the daemon? It's unclear whether pressing Enter is expected.

Yuncun:
**15. Loops inside a composite can't address themselves.** A goal loop created with `--into <composite>` ran `graphcode node memo <project> <its-own-id> …` as its prompt instructs and got *no loop <id> in this graph*; `graphcode status <project>` lists only the composite. Same for `node refine`, `node delete`, `edge create` from outside. So sub-graph children have no memo/playbook path and can only be managed in the app. Either accept the child id at the project path, or document a sub-graph path form the CLI accepts.

scgopi:
Thanks @Yuncun — all three items investigated against the source (unchanged from 0.1.56 in these areas), and every one reproduces in code. Here is what is actually happening for each.

---

## 13. `--skip-unchanged` deadlock — confirmed, and it's worse than a missing warning

[...]

PR #222, opened 2026-08-30
Fix --skip-unchanged deadlocking a goal loop that is the only writer of its tree (#217 item 13)

scgopi:
## The deadlock

With `--skip-unchanged`, every goal poll captured a workspace fingerprint (`git rev-parse HEAD` + `git status --porcelain` hashed) and returned early if it matched the one recorded at the last failing run — no predicate re-run, no failure relayed (GraphStore.swift:2053-2062 at 0.1.56).

For a loop that is the *only writer of its tree*, that is a fixed point: the loop only writes once woken, and waking was gated on a tree change. After a turn that ends without moving the tree, every poll skips forever — the ~60 min silence from issue #217 item 13.

## The fix

- **The skip now only applies while the session is busy.** That is the case its optimisation is for: the agent is mid-turn, its next write is what would change the tree, and re-running an expensive predicate buys the same failing answer. A missing fingerprint still falls through to a real run.
- **Idle plus unchanged is wake-worthy — once per frozen tree.** On the first poll that finds an idle session on an already-failed tree, the predicate runs again (the only path on which an external watcher's change is ever seen) and the failure is re-delivered even when the tail is identical. A marker (`reawakenedFingerprints`) then holds the skip until the tree moves; a new failing run at a changed fingerprint makes one more wake available. Bounded: one extra predicate run and one extra relay per frozen tree, never a relay per poll — the unbounded agent-turn spend the failure-tail dedup exists to prevent stays prevented.
- **Presence-less sessions stay skipped, deliberately.** The relay refuses to tell a session it cannot see idle, so falling through for a nil presence would only buy the predicate's price for a wake that cannot land; the contract is documented at the guard and pinned by a test. Such a loop's exits are its stall bound and its human.
- **Help text corrected.** It previously said to use the flag "only for predicates that depend on the tree" — the exact case that deadlocks. Now it states the busy/idle contract and the one-notice-per-frozen-tree bound.
- **Create- and update-time warning.** `node create` and `node update --skip-unchanged true` print the advice on stderr (stdout stays the rendered graph) via `GraphcodeCommand.createWarnings(for:)` / `.updateWarnings(for:currentNode:)`. The update path is best-effort: a node not visible at the top level (a sub-graph child — item 15) warns nobody rather than warning wrongly.

## Tests

Six tests in `PredicateFeedbackTests` cover the new behaviour:

- `anIdleLoopOnAnUnchangedTreeIsWokenOnceNotEveryPoll` — the item 13 regression, encoding the bound: wake once, then quiet (predicate runs and deliveries both stop).
- `aTreeChangeBuysOneMoreWakeOnTheNextFreeze` — the bound is per fingerprint, not per loop lifetime.
- `aPresencelessSessionStaysSkippedRatherThanPayingForAWakeThatCannotLand` — nil presence takes the skip.
- `anIdleLoopOnAnUnchangedTreeStillNoticesThePredicatePassing` — a CI-style predicate going green on an unchanged tree still resolves the node, on the wake poll.
- `theCreateWarningFiresOnlyForSkipUnchangedPairedWithAPredicate` and `theUpdateWarningFiresWhenSkipUnchangedIsTurnedOnForAPredicatedGoalLoop` — the warnings gate exactly on goal + predicate + flag.

Full suite: 1299 tests passed. swiftlint 0 errors; swift-format strict clean on the touched files.

Refs #217 (item 13).

### DIFF
--- a/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
+++ b/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
@@ -123,10 +123,13 @@ public enum GraphcodeCommand: Equatable, Sendable {
                            many tokens spent (input + output). Reported, never
                            estimated — a loop whose backend reports nothing is never
                            stopped by a budget
-      --skip-unchanged     for --type goal: don't re-run the predicate while HEAD and
-                           the dirty file list are unchanged since its last failure.
-                           Only for predicates that depend on the tree — one that
-                           watches CI or a deploy would never be re-asked
+      --skip-unchanged     for --type goal: while the session is busy, don't re-run
+                           the predicate while HEAD and the dirty file list are
+                           unchanged since its last failure. Once the session goes
+                           idle the predicate is re-run and a failure re-delivered
+                           even on an unchanged tree — the loop is the only writer
+                           of its own tree, so waiting on a change would wait on
+                           the loop itself
 
     UPDATE OPTIONS (node update; pass only what changes)
       --goal, --predicate, --prompt, --check, --model, --metric, --direction as above
@@ -741,6 +744,25 @@ extension GraphcodeCommand {
       return "incomplete loop: a turn-based node needs --check, a goal-based one --goal, "
         + "a time-based one --prompt, and the backend must be able to host that type"
     }
+  }
+
+  /// Advice printed at `node create` time for the flag combination a first-time user
+  /// reached for and got stranded by (issue #217 item 13): `--skip-unchanged` on a goal
+  /// loop with a predicate. The flag's name invites reading it as "the daemon handles
+  /// idle polls", when what it does is skip re-runs while the session is busy — the
+  /// loop itself is the only writer of the tree the skip was watching. The poller now
+  /// wakes an idle loop on an unchanged tree, so this is advisory rather than a
+  /// refusal, but the contract is worth saying out loud where the flag is typed.
+  public static func createWarnings(for draft: NodeDraft) -> [String] {
+    guard draft.loopType == .goalBased, let goal = draft.goal,
+      goal.skipsUnchangedWorkspace, goal.effectivePredicate != nil
+    else { return [] }
+    return [
+      "warning: --skip-unchanged only spares the predicate while the loop's session is "
+        + "busy; once the session goes idle the predicate is re-run and a failure "
+        + "re-delivered even on an unchanged tree, because the loop is the only writer "
+        + "of its own tree"
+    ]
   }
 }
 
--- a/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
+++ b/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
@@ -125,11 +125,11 @@ public enum GraphcodeCommand: Equatable, Sendable {
                            stopped by a budget
       --skip-unchanged     for --type goal: while the session is busy, don't re-run
                            the predicate while HEAD and the dirty file list are
-                           unchanged since its last failure. Once the session goes
-                           idle the predicate is re-run and a failure re-delivered
-                           even on an unchanged tree — the loop is the only writer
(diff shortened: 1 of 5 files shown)


scgopi:
## Review verdict: ✅ Approve

Reviewed against issue #217 item 13 and verified on the PR head (`78ed83a`) from a worktree. (GitHub won't accept a formal self-approval from this account, so the verdict is posted as a comment — the review itself is full-strength.)

## Correctness vs the reported deadlock

The pre-fix skip (`fingerprint matches → return`, applied on every poll) was a fixed point for a loop that is the only writer of its tree: it only writes once woken, and waking was gated on a tree change. The fix resolves both halves of that:

| Path | Behaviour | Verdict |
|---|---|---|
| Busy + unchanged tree | skip retained (skip predicate re-run) | ✅ the case the optimisation is for |
| Idle + unchanged tree | predicate re-runs; `lastPredicateFeedback` cleared so the identical-tail dedup in `relayPredicateFailure` (GraphStore.swift:2155) can't suppress the wake | ✅ un-strands item 13 |
| Idle + unchanged + CI goes green | predicate re-runs → node resolves | ✅ covered by test |
| `.absent` / `.unknown` / `.awaitingInput` presence | still skips | ✅ same as pre-fix — the fix is a strict superset of wake paths, nothing removed |

Details that check out:

- The `presence == .idle` gate exactly mirrors `relayPredicateFailure`'s own deliverability gate (GraphStore.swift:2162), so the wake only opens when a relay could actually be delivered.
- `createWarnings`' `effectivePredicate != nil` guard matches the poller's `hasPredicate` logic — the warning fires only when the flag is paired with a live predicate (a blank predicate never arms that half of the poller).
- Re-delivery every poll while idle+unchanged+failing is the right call, not an accident: deliver-once-per-idle would re-strand (loop wakes, changes nothing, dedup suppresses the next wake). The help text and the create-time warning state this contract.
- Warning goes to stderr; stdout stays the rendered graph. Also fires on the `--into` composite path (same draft) — correct.
- Docs fix is right: the old help recommended exactly the case that deadlocked.

One non-blocking nit: the new idle tests exercise the `node.presence` fallback (no `onReadPresence` wired), not the production `onReadPresence` path. Production wiring is a thin pass-through of the same `Presence` value, so risk is low — fine to leave.

## Verification on the PR branch

| Check | Result |
|---|---|
| Full test suite (`xcodebuild test`, worktree at `78ed83a`) | ✅ **1296 tests passed** (1293 before, +3) |
| All 3 new tests ran (not silently skipped) | ✅ confirmed in the log by name |
| `swiftlint lint` errors | ✅ 0 |
| `swift format lint --strict` on the 5 touched files | ✅ clean |

Fixes the bug, no regressions found. Approving.

scgopi:
All three findings addressed in 88cf3e7:

1. **Wake-spam** — the re-delivery is now spent once per fingerprint (`reawakenedFingerprints`): the idle+unchanged poll re-runs the predicate and re-delivers once, then the skip holds until the tree moves; a failing run at a new fingerprint makes one more wake available. `anIdleLoopOnAnUnchangedTreeIsWokenOnceNotEveryPoll` now asserts 4 polls → 2 predicate runs, 2 deliveries; `aTreeChangeBuysOneMoreWakeOnTheNextFreeze` pins the per-fingerprint (not per-lifetime) semantics.
2. **Presence nil** — stays skipped, with the reason documented at the guard: the relay only tells a session it can see idle, so falling through would pay the predicate's price for a wake that cannot land. Pinned by `aPresencelessSessionStaysSkippedRatherThanPayingForAWakeThatCannotLand`.
3. **node update warning** — `GraphcodeCommand.updateWarnings(for:currentNode:)` prints the same advice when `--skip-unchanged true` lands on a goal loop with a predicate (judging the predicate as the update leaves it), using the graph the update path already loads. Best-effort: an unresolvable node warns nobody rather than wrongly.

Suite at 1299 passing; lint/format clean.

PR #229, opened 2026-08-31
Mailboard: a shared, unaddressed board for loops (beta-ramped)

scgopi:
## What this adds

`node send` and edges are **addressed** — a loop must already know a peer's id, and the daemon routes to that one peer. The Mailboard is the ambient counterpart: one append-only board per graph that any loop can **post to for whoever comes next** — a decision made, a dead end hit, a claim staked — and read back with one command, with no wiring and no ids. Posts survive their authors: a loop created after the writer is gone still finds the note at its next wake.

## The verbs

```sh
graphcode mailboard post <path> [--topic <t>] <note>   # leave something behind
graphcode mailboard sync <path>                        # read your unread (and mark them read)
graphcode mailboard list <path>                        # peek without moving your cursor
graphcode mailboard watch <path> [--topic <t>]         # hear new posts while you live
```

Attribution rides `ZMX_SESSION` exactly like `node send`; posts from a human's shell show as "a human".

## How it works

- **`MailboardKit` module** (new Tuist target, Foundation-only): the post/watch model, caps (1 KiB body, 200 posts, oldest pruned), and the unread arithmetic every surface shares.
- **Board state rides `LoopGraph`**: one writer (the daemon), persisted beside the graph file, snapshot in every `.graphChanged` — which is the CLI's read path, so no second read protocol. The global graph at `graphcode://global` becomes a cross-project board for free. Absent-while-empty encoding keeps untouched graph files byte-identical, and old files decode with an empty board.
- **Watcher wakes** ride `deliverAdHocMessage`'s follow-up semantics: typed into a live idle session, queued behind a busy turn, staged to memory otherwise. The post itself is the durable half; the ding is best-effort.
- **Discovery**: the generated briefing teaches the verbs (only while the feature is on) and the wake digest reminds relaunching loops to check the board before redoing work a predecessor may have posted about.

## The ramp

**Beta only, kill-switchable.** `FeatureRamps.mailboard` ships `beta: 100 / stable: 0`; the app resolves the ramp into `mailboardEnabled` in `~/.graphcode/settings.json` — the one bit the daemon (which cannot see ramps) enforces on every command, the briefing section, and the digest line. An explicit Settings choice outranks the ramp forever; the daemon refuses board commands out loud with a pointer to the setting while it is off.

## Verification

- New tests: `MailboardTests` (store handlers, cursors, watcher wakes, pruning, Codable back-compat), `MailboardCommandTests` (CLI parsing/attribution), `SettingsMailboardTests` + `FeatureRampsTests.mailboardShipsBetaOnAndStableOff`.
- Full suite: **1316 tests / 143 suites pass**; GraphcodeKit and app schemes build clean.


scgopi:
Rename the Mailboard to Artifactory

scgopi:
Every shared communication should be stored inthe artifactory as well

scgopi:
Both comments addressed (f613a66), plus the deletion question you raised:

1. **Renamed Mailboard → Artifactory** end to end: `ArtifactoryKit` module, `ArtifactoryPost`/`ArtifactoryWatch`, `artifactoryEnabled` setting + ramp key, `graphcode artifactory post|sync|list|watch` verbs, briefing and wake-digest wording.

2. **Every shared communication is now stored in the artifactory**: `node send` (immediate and follow-up) lands as a `direct` record, delivered message-edge deliveries as `direct`, handoffs — payload included — as `handoff`. Mirroring is record-only (it never rings watchers, so a busy graph doesn't double-deliver), and undelivered edge messages are deliberately not recorded: the artifactory is a record of what was actually said.

3. **Deleting a loop deletes the artifactory posts it authored** along with the edges/session/memory teardown delete already does — spawned descendants included. Posts where the loop was only the recipient stay (the other side's record), and `node stop` keeps everything.

Full suite: 1323 tests / 143 suites pass, including 8 new tests for mirroring and deletion cleanup.

scgopi:
Independent review addressed in full (833e4e2), including the merge of the stale base:

**Majors**
1. **Composite gating** — the gate now forwards into both sub-graph store construction sites; a piloted composite's workers post/sync/watch and mirror exactly like any loop, with a test pinning it. `nil` still means off — forwarding was the fix, not a nil-means-on reading.
2. **Stale base** — `origin/main` merged (was based on 1ecdec1). Version is now **0.1.58-beta1 / build 222**, past main's 221 rather than colliding at 217. The one Linux-CI-sensitive call (`Date.formatted`) is replaced by a fixed-`dateFormat` `DateFormatter` with a pinned locale.

**Minors** — 3 (truncation now really ≤1024), 4 (race comment states the race), 5 (dead `direct`/`handoff` watches documented in help + briefing), 6 (MAILBOARD/grammar), 7 (imports fixed; all touched dirs pass `swift format --strict`), 8 (imported loops start with a clean cursor; watch travels as a preference).

**Nits** — 9 (foreign sender reads "an outside loop"), 10 (off-briefing is byte-identical to pre-Artifactory), 12 (idempotent `watch --off`, help spacing, topic doc sentence). Nit 11 accepted: the ack pattern is racy by design and the body is what matters.

**Tests** — 10 added for the review's holes; full suite **1371 tests / 144 suites pass**.

scgopi:
Read-side review addressed in full (79087bb) — all findings, including the optional polish:

1. **Status line on the no-loops path** — a board with human posts outlives every loop; "no loops yet" no longer hides it.
2. **`--search --json` composes** — the filter threads into the JSON; the combination can't quietly return the whole board.
3. **`read` rejects non-positive ids** at parse (`-7` is a typo, not a post); leftover-word looseness left as-is per the parser's documented positional contract.
4. **Sync flag precedence documented** in help (`--json > --mark > --headlines`; cursor advances regardless).
5. **JSON dates are ISO-8601**, pinned by an explicit decoding-strategy test so the wire format can't drift silently.
6. **"Unread for you" claimed only for a reader this graph knows** — foreign/stale ids get the plain count, matching the daemon's sync refusal.
7. Headline truncation flattens newlines; "marked read up to #0" is now a sentence; the "a artifactory" typo is gone.

New tests cover every named gap, including the exact-80 headline boundary and the nodes-empty+human-posts state. Full suite: **1384 tests / 144 suites pass**; `swift format --strict` clean.

PR #232, opened 2026-09-01
An animation for the Artifactory

scgopi:
A designed HTML figure explaining the Artifactory (#229), in the house style of the
existing `marketing/animations/` figures — 1200×675, same palette and type as
`loops-talk`, stepped through `window.__seek` and rendered by `./build.sh artifactory`.

**The story, four beats, the last one longest.** `node send` and edges are addressed:
the figure opens on a drawn wire between two loops and a command carrying a uuid the
sender had to already know. Then Fixer posts to the board — the pip travels to the
Artifactory along *no wire at all*, because nothing was wired and nobody was named.
Fixer resolves, is deleted, and the lane closes over it while its post's byline becomes
`Fixer (deleted)`. Finally Auditor, created three days later and wired to nothing, sees
the unread count in `status` and reads the note with one `sync`. That last beat gets a
third of the runtime, because it is the claim: notes outlive their authors.

**Nothing on screen is invented.** Every command and every line of output is what the
merged code prints — `renderPosted`, `renderArtifactory`'s
`<project> artifactory, unread: N post` header and
`#<id> (topic) from <author> at <stamp> — <body>` rows, `renderArtifactoryStatusLine`,
and the empty-board line — with the byline taken from
`ArtifactoryPost.withAuthorDeleted`. A new loop's `lastArtifactoryRead` is `nil`, so its
first sync genuinely does see the whole board.

| | |
|---|---|
| Duration | 30.0s, 25fps capture / 12.5fps GIF |
| GIF | **766,186 bytes** (~748 KiB), inside the ~900 KB README/site budget |
| MP4 | 608,805 bytes, 1600px CRF 18 |
| Deterministic | `./build.sh artifactory` twice → identical sha256 on all three outputs |

**Two commits, and the first needs a decision.** `marketing/animations/` has never been
tracked — not ignored, just never committed — so a figure landing on its own would be a
GIF nobody else could rebuild or check. The first commit brings the four renderer
scripts (`build.sh`, `render.mjs`, `probe.mjs`, `package.json`/lock) along, plus a
`.gitignore` for `frames/` and `node_modules`. The other nine local figures are *not*
included, which leaves one wart: `./build.sh` with no arguments still names them in its
default list and will fail on a fresh clone. `./build.sh artifactory` works. Say the
word if you would rather the whole directory came across, or none of it.

https://claude.ai/code/session_011b4JC1STxSMfbx7fJQv5wz


scgopi:
Closing — marketing/ is kept local and is not intended to merge.

PR #263, opened 2026-09-03
Rename a loop when its session is renamed (#261)

### LINKED ISSUE #261: A session renamed with /rename should rename its loop
**Is your feature request related to a problem?**

Renaming a Claude Code session from inside it (`/rename hello`) changes the session's name, but the loop's card in GraphCode keeps whatever title it was created with. The two names then disagree, and the sidebar keeps showing a name the human has already replaced.

The loop title is the one field a human is expected to change after the fact — it's written before the work exists, and what the loop turns out to be doing is only known once it's running. `/rename` is a human saying exactly that, in the session, and GraphCode ignores it.

**Describe the solution you'd like**

When a session's title changes, rename the node to match.

Claude Code writes an explicit record to the transcript GraphCode already tails:

```json
{"type":"custom-title","customTitle":"hello","sessionId":"a825abc7-..."}
```

This is written only by `/rename`, so it is an unambiguous statement of human intent — distinct from the `agent-name` record, which carries an auto-assigned name (`angleReuse2`) and appears in sessions nobody renamed. Only `custom-title` should be read.

**Scope**

- Session → loop only. An app-side rename should not be pushed back into the session, and must not be clobbered by the next poll.
- Claude Code only. The other three backends have no equivalent record; they should report nothing rather than a guess.

**Additional context**

The plumbing already exists: `GraphCommand.renameNode`, `ClaudeSessionLog.transcript(forSessionID:)`, and the 15s `pollPresence` tick that already opens the same file for the summary rail.

scgopi:
Fixes #261.

Renaming a session from inside it (`/rename hello`) left the loop's card showing whatever title it was created with. The title is the one field a loop is expected to outgrow — it is written before the work exists — so the two names disagreed at exactly the moment the human had just said what the loop is.

## Where the answer comes from

Claude Code writes it down, in the transcript GraphCode already knows how to find:

```json
{"type":"custom-title","customTitle":"hello","sessionId":"a825abc7-…"}
```

It is a standalone record, re-emitted at every checkpoint rather than once at the rename, so a tail read always carries the current one and the last one wins. Read on the existing 15s presence poll, applied through the same `renameNode` a human's rename goes through — so a blank title is refused and everything a rename deliberately leaves alone (id, session, edges, layout) stays left alone.

## Three things this is careful about

**`custom-title` only.** Every session also carries an `agent-name` record, and it is not the same thing: it holds a name the CLI assigned itself (`angleReuse2`) and is present in sessions nobody has renamed. Reading it would retitle most of a graph to strings no human ever typed — a far worse bug than the one being fixed. `anAgentNameIsNotARename` pins this.

**Once per change, not once per sight.** `LoopNode.sessionTitle` records what the session was last *observed* to be called, which is not what the node is called — the two part company the moment somebody renames the card from the sidebar, and that gap is what keeps the card's name theirs instead of having it overwritten on the next tick. It is persisted for the same reason: a daemon restart must not make an applied rename look new again.

**Cost.** `summarisesLoops` is off by default, so on a default install nothing was reading transcripts at all; ungated, this would have added a 512KB tail read per loop every fifteen seconds to every install there is, to re-read a name that changes when a human types six characters. Gated on `TranscriptFreshness` — its own instance, not `shared`, because `hasChanged` is consuming and the two readers would otherwise take turns to work — a quiet loop costs one `stat`.

## Deliberately out of scope

- **Remote loops.** Reading one costs an ssh probe per node per poll. Worth it for a rail that moves every few seconds, not for a name that changes once in a loop's life. Reported as "nothing said" rather than bought at that price.
- **The other three backends.** None of Copilot CLI, Codex, or OpenCode has a rename that leaves a record behind, so they return `nil` rather than a guess.
- **The reverse direction.** An app-side rename is not pushed into the session.

## Regression surface

| Risk | Guard |
|---|---|
| Clobbering a human's title every 15s | `sessionTitle` change-detection — `aRenameFromTheSidebarSurvivesTheNextPoll` |
| Renaming loops from `agent-name` | Record type matched exactly — `anAgentNameIsNotARename` |
| The poll starting to write graph.json every tick | Persist only when a rename actually landed; the existing `thePollMustNotPersistTheGraph` still passes |
| Starving the summary rail's freshness gate | Separate `TranscriptFreshness` instance; `SummaryRailTests` unchanged and green |
| Old graphs failing to decode | `decodeIfPresent`, matching every other field on `LoopNode` |

## Verification

| Check | Result |
|---|---|
| `xcodebuild test` | ✅ 1534 tests / 159 suites, up 11 |
| `graphcode` / `graphcoded` / `graphcode-cli` builds | ✅ all three succeed |
| `swiftlint lint` | ✅ 0 errors |
| `swift format lint --strict` | ✅ clean |

https://claude.ai/code/session_011hEyiK8W9XW4Tq6je98kTX

### DIFF
--- a/GraphcodeKit/Sources/Domain/LoopNode.swift
+++ b/GraphcodeKit/Sources/Domain/LoopNode.swift
@@ -126,6 +126,17 @@ public struct LoopNode: Identifiable, Codable, Equatable, Sendable {
   /// `nil` until the summary producer is switched on in Settings, which is where it stays
   /// for anyone who never turns the experiment on.
   public var summary: LoopSummary?
+  /// What this node's session last said it was called — `/rename` inside Claude Code.
+  ///
+  /// Recorded so the rename fires **once per change** rather than once per poll. Without
+  /// it, the 15s tick would re-apply the session's name forever, and renaming a loop from
+  /// the sidebar would silently undo itself a few seconds later. Persisted for the same
+  /// reason: a daemon restart must not make an already-applied rename look new again.
+  ///
+  /// It is the last *observed* title, not the node's title — the two differ the moment a
+  /// human renames the card afterwards, and that difference is what keeps the card's name
+  /// theirs.
+  public var sessionTitle: String?
   /// The same run, drawn — a Mermaid flowchart or a table, composed once per finished pass
   /// and rendered natively by the rail (`SummaryBoard`).
   ///
@@ -213,6 +224,7 @@ public struct LoopNode: Identifiable, Codable, Equatable, Sendable {
     usage: UsageSample? = nil,
     activity: String? = nil,
     summary: LoopSummary? = nil,
+    sessionTitle: String? = nil,
     board: SummaryBoard? = nil,
     presence: PresenceReading? = nil,
     metricHistory: [MetricSample] = [],
@@ -242,6 +254,7 @@ public struct LoopNode: Identifiable, Codable, Equatable, Sendable {
     self.usage = usage
     self.activity = activity
     self.summary = summary
+    self.sessionTitle = sessionTitle
     self.board = board
     self.presence = presence
     self.metricHistory = metricHistory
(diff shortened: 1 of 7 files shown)


scgopi:
Don't merge the PR yet, need to see if it is low risk

PR #274, opened 2026-09-04
Show loop-to-loop discussion on the board (#273)

### LINKED ISSUE #273: Loop-to-loop discussion is invisible on the board: 'direct' replies mirror as records and hide in a rollup that forgets it was opened
Loop-to-loop replies are mirrored as `record` posts, and the rail's board hides every one of them behind a one-line rollup that forgets it was opened. When two loops actually discuss something — a root-cause correction, an answer to a review question — the supervisor watching the board sees nothing.

## What happened

Two loops (`TripleAgentSpawn`, `IssueTriage`) worked through the diagnosis that became #272 entirely over `graphcode node send`. Five posts landed on the board; one was visible.

| Post | Topic | Kind | Content | Visible in the rail |
|---|---|---|---|---|
| #69 | `direct` | `record` | Reported the codex-pane bug + a test-coverage question | ❌ rollup |
| #70 | `issue-272` | `note` | "Filed #272 and CLAIMING it" | ✅ board list |
| #71 | `direct` | `record` | Heads-up to the Release loop about 0.1.61-beta1 | ❌ rollup |
| #72 | `direct` | `record` | Root-cause **correction**: truncation is not the mechanism, `zmx run` sets `.command = null` | ❌ rollup |
| #73 | `direct` | `record` | Correction accepted + verified, answered the test-coverage question | ❌ rollup |

Store-wide on this project right now: 50 notes, 23 records.

The two posts carrying the actual technical substance (#72, #73) are the two least visible.

## Why

`graphcode/Sources/Features/LoopWorkspace/ArtifactorySection.swift`

| Line | Behavior |
|---|---|
| `:19` `notes(in:)` | Main list filters to `kind == .note` |
| `:26` `records(in:)` | Direct messages and handoffs split off — "receipts for deliveries that already happened, not something written to be read here" |
| `:34` `unreadNoteCount` | Records excluded from the badge — "a badge counting mail nobody is being shown is a badge that cannot be cleared" |
| `:62` | Records-expanded state is **local and unpersisted** |
| `:238-269` | `recordsRollup` renders "N message records", expanding to the last 8; older ones have no affordance at all |

Each decision is defensible on its own. Together they mean a channel loops genuinely use for discussion is one a supervisor never reads — which the section's own doc comment names as the failure mode it exists to prevent:

> a coordination channel a supervisor never sees is the failure mode, not a missing convenience

## The real problem

The `direct` topic does double duty. It carries both:

- **receipts** — "handoff delivered", "message staged" — which genuinely are noise, and
- **discussion** — corrections, answers, review findings — which is exactly what the board is for.

`kind == .record` is applied on delivery mechanism, not on content, so the two are indistinguishable downstream.

## Possible directions

1. **Persist the expanded state.** Smallest fix. `:62`'s local `@State` becomes an `AppStorage` key like the section's own fold. Doesn't fix the badge or the >8 cutoff.
2. **Count records in the badge** once they can be shown and cleared, removing the reason `:34` excludes them.
3. **Separate receipt from discussion.** A delivery receipt is a record; a message a loop *wrote* is a note. This is the fix that matches the stated intent — the mirrored body is already the full message text, so the information to make the split is present at mirror time.
4. **Drop the last-8 cutoff** in the rollup, or give "N earlier" a way to open.

(3) plus (2) is my read of what the section's own reasoning is asking for, but the call is a design one.

## Repro

1. Two loops in a project; from one, `graphcode node send <other-id> <a substantive message>`.
2. Have the target reply the same way.
3. Open either loop's rail. The exchange is not in the board list; it is inside a collapsed "2 message records" line, and there is no unread badge for it.
4. Expand it, navigate away, come back — it is collapsed again.

Found while verifying #271 and diagnosing #272.

scgopi:
Closes #273.

## The problem

The rail split the Artifactory on `ArtifactoryPost.Kind`, and that field is a *budget*: mirrored traffic is written as `.record` so a talkative graph prunes on its own quota instead of evicting notes. It was also being read as "is there anything in here to read", which it never was — a `node send` carries every word its sender typed and mirrors as a `.record` all the same. So the rail hid it, the `N NEW` badge skipped it, and the rollup forgot it had ever been opened.

Measured on this repo's own board while the issue was filed: the two posts carrying the actual technical substance (a root-cause correction and its acceptance) were the two least visible.

## The fix

`ArtifactoryPost.wasWritten` splits the axes.

| Mirrored traffic | `wasWritten` | Where it lands |
|---|---|---|
| `node send` | ✅ | a row on the board, counted by the badge |
| `.message` / `.handoff` edge with a template or script payload | ✅ | a row on the board |
| edge with no payload transform (`"Author finished."`) | ❌ | the receipts rollup |
| hand-off nudge with no payload | ❌ | the receipts rollup |

Direction (3) + (2) from the issue, with one deliberate departure: the issue proposes making a written message a `.note`. That would put `node send` traffic in the note budget and reinstate the regression `Kind`'s own doc comment records — two hundred sends evicting every note on the board. A third `Kind` case would fix the budget but hand an older build a raw value it cannot decode, and `loadGraph` returning `nil` costs a project its whole graph. A flag beside the kind keeps both properties, and old boards decode with their records still folded — exactly where they are today.

Also in here, per the issue's other two directions:

- **:62 persists.** The rollup's open state moved from local `@State` to `@AppStorage`, beside the section's own fold.
- **The `N earlier` cutoff opens.** It was a line that named something and did nothing; it is now `N earlier — show all`. The scroll box already clamps its own height, so opening the rest costs the rail nothing.
- The composer moved into a trailing `extension`, buying back the `type_body_length` lines this change spends. Pure move, no logic.

## Verification

From `worktrees/fix-273-visible-messages`, per the `worktree-build-check` skill:

| Check | Result |
|---|---|
| `xcodebuild -scheme graphcode build` | ✅ 0 errors |
| `xcodebuild -scheme graphcode test` | ✅ 1548 tests, 160 suites, passed |
| `-scheme graphcode-cli` / `-scheme graphcoded` build | ✅ 0 errors (GraphcodeKit changed) |
| `swiftlint lint` | ✅ 0 errors |
| `swift format lint --strict` | ✅ clean |

Seven new tests, each confirmed present in the run: the written/receipt split at all three mirror sites, that written messages still prune on the record budget, that a pre-split board decodes its records as receipts, and that the badge both counts a written message and is cleared by leaving the workspace — the property the section's doc comment insists on.

https://claude.ai/code/session_01Gz9hh5iq8MNpr9QSCjkbnt

### DIFF
--- a/ArtifactoryKit/Sources/Artifactory.swift
+++ b/ArtifactoryKit/Sources/Artifactory.swift
@@ -116,9 +116,12 @@ public struct ArtifactoryPost: Codable, Equatable, Identifiable, Sendable {
     authorID = try container.decodeIfPresent(UUID.self, forKey: .authorID)
     author = try container.decode(String.self, forKey: .author)
     topic = try container.decodeIfPresent(String.self, forKey: .topic)
-    body = try container.decode(String.self, forKey: .body)
     kind = try container.decodeIfPresent(Kind.self, forKey: .kind) ?? .note
-    wasWritten = try container.decodeIfPresent(Bool.self, forKey: .wasWritten) ?? (kind == .note)
+    let body = try container.decode(String.self, forKey: .body)
+    self.body = body
+    wasWritten =
+      try container.decodeIfPresent(Bool.self, forKey: .wasWritten)
+      ?? (kind == .note || !Artifactory.readsAsADeliveryReceipt(body))
   }
 
   /// The same post with its author's handle gone — what deleting a loop leaves behind.
@@ -164,6 +167,26 @@ public enum Artifactory {
   /// enough that a loop joining mid-flight can see what was recently said, not so
   /// many that the graph's chatter becomes the board.
   public static let maxRecords = 50
+
+  /// Whether a record saved before `wasWritten` existed is one of the two lines the
+  /// daemon generates itself, rather than something a loop said.
+  ///
+  /// Boards written before the split carry no flag, and a default of "receipt" would
+  /// leave every conversation already on them exactly as buried as #273 found it — a
+  /// fix that only helps graphs created after it shipped. There is no other signal left
+  /// on those posts, so this reads the two shapes the mirror produces when an edge fires
+  /// with nothing in it (`MessageBus.messageText`'s `.none`, and a hand-off with no
+  /// payload). Everything else on those topics is a `node send` or a payload, which is
+  /// somebody talking.
+  ///
+  /// Deliberately narrow. A written message that happens to end "… finished." is read as
+  /// a receipt and stays folded, which is where it already was; the opposite mistake
+  /// would put the daemon's own bookkeeping in front of a reader as though a loop had
+  /// said it. Only posts decoded without the flag are ever asked.
+  public static func readsAsADeliveryReceipt(_ body: String) -> Bool {
+    body.hasSuffix(" finished.")
+      || body.hasSuffix(" finished and handed its work off to you.")
+  }
 
   /// The id the next post gets. Maximum-plus-one, never count-plus-one: pruning
   /// removes the oldest posts, and reusing their ids would make unread cursors
--- a/ArtifactoryKit/Sources/Artifactory.swift
+++ b/ArtifactoryKit/Sources/Artifactory.swift
@@ -55,6 +55,26 @@ public struct ArtifactoryPost: Codable, Equatable, Identifiable, Sendable {
   /// own quota, where everything on the board was a note.
   public let kind: Kind
 
+  /// Whether a loop *wrote* this, or the board is only noting that a delivery happened.
+  ///
+  /// A separate axis from `kind` on purpose, and the two were conflated once. `kind`
+  /// answers "whose budget prunes this" — mirrored traffic must have its own quota or a
+  /// talkative graph evicts every note. It was also read as "is there anything in here
+  /// to read", which it never was: `node send` mirrors as a `.record` because that is
+  /// the budget it belongs to, and the whole text a loop typed rode along inside it.
+  /// Every surface then folded it away, so two loops correcting each other's diagnosis
+  /// held the conversation somewhere no supervisor ever looked (#273).
+  ///
+  /// A hand-off nudge carrying no payload really is a receipt. The text of a `node send`
+  /// is not, and this is the bit that says so.
+  ///
+  /// Absent from boards saved before the split, where a record was only ever a receipt —
+  /// which is also why it rides beside `kind` rather than becoming a third case of it:
+  /// an older build decoding a newer board must not meet a raw value it has never heard
+  /// of, and losing a project's whole graph to a rolled-back beta is a steep price for a
+  /// tidier enum.
+  public let wasWritten: Bool
+
   /// Cached formatter for CLI rendering — one `DateFormatter` per process rather than
   /// per post, and a fixed `dateFormat` with a pinned locale rather than
   /// `Date.formatted` or named `DateFormatter.Style` cases, neither of which is
@@ -66,9 +86,11 @@ public struct ArtifactoryPost: Codable, Equatable, Identifiable, Sendable {
     return formatter
   }()
 
+  /// `wasWritten` defaults to what the kind implies: a note is always somebody speaking,
+  /// and a record is a receipt unless its caller says otherwise.
   public init(
     id: Int, at: Date, authorID: UUID?, author: String, topic: String?, body: String,
-    kind: Kind = .note
+    kind: Kind = .note, wasWritten: Bool? = nil
   ) {
     self.id = id
     self.at = at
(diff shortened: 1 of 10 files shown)


scgopi:
## Review of #274 (high effort, adversarial)

Reviewed `fix/273-visible-messages` (`4a06fcd`, `62134d7`, `e86a2ad`) against `main`, reading through `GraphStore`, `MessageBus`, `Artifactory` and the rail.

**The core logic is sound.** `wasWritten` is classified correctly at all three mirror sites (`payloadTransform != .none`, `payload != nil`, `true` for `node send`), the decode default is backward-compatible and scoped to legacy posts only, `withAuthorDeleted` forwards the flag, every caller of the renamed `notes`/`records`/`unreadNoteCount` API was updated, and `wakeArtifactoryWatchers` still passes `mirror: false` so the badge cannot feed itself.

Five findings, one of which should block.

| # | Location | Sev | Issue |
|---|---|---|---|
| 1 | `GraphStore.swift:2370` | 🟠 **Medium — blocking** | Receipts can silently evict the written conversation this PR exists to show |
| 2 | `Artifactory.swift:186` | 🟡 Low | Receipt literals duplicated across modules with nothing tying them together |
| 3 | `ArtifactorySection.swift:303` | 🟡 Low | "N earlier — show all" is visually inert once the scroll box is at `maxHeight` |
| 4 | `ArtifactorySection.swift:85` | 🟡 Low | `showsAllReceipts` is one-way, contradicting its own doc comment |
| 5 | `LoopWorkspaceFeature.swift:328` | 🟡 Low | A folded board now accrues an unclearable, unbounded `N NEW` badge |

### 1 — Receipts evict written messages (`GraphStore.swift:2370`)

`Artifactory.pruned` keeps the newest 50 `.record` posts regardless of `wasWritten`, and `drainPendingMessages` records a receipt on **every** delivery with no cycle-re-entry exemption — unlike the handoff path at `:2462`, which skips `isCycleReentry`. `reenterCycle` resets `fireCount = 0` for every edge between cycle members, so an unguarded `.none`-transform `.message` edge on a cycle mirrors a fresh `"@X: Y finished."` receipt every pass.

**Failure scenario:** two loops work through a root cause via `node send`; a cycling graph then runs 50 more passes. The record budget is now 100% receipts, the written discussion is evicted, and the section renders "50 delivery receipts" and nothing else.

Before this PR both halves were hidden, so the eviction was invisible. **After it, the same eviction deletes the board's visible content** — which makes this a regression the PR introduces into its own feature.

Fix direction: exempt payload-less receipts from the record budget, or give them a smaller sub-quota beneath it.

### 2 — Cross-module literal duplication (`Artifactory.swift:186`)

`readsAsADeliveryReceipt` matches `" finished."` and `" finished and handed its work off to you."`, but those strings are generated in `GraphcodeKit` — `MessageBus.messageText` case `.none`, and `drainPendingHandoffDeliveries`'s `parts.append(...)`. The literal is duplicated across a module boundary rather than shared.

Reword either string, even its punctuation, and legacy boards silently reclassify every payload-less receipt as written: the rollup empties, every stale `"Author finished."` takes a row, and `N NEW` jumps by the whole backlog. **No compile error and no failing test** — the new tests hardcode the same literals independently.

Suggest hoisting the two suffixes into `ArtifactoryKit` constants that `MessageBus` and `drainPendingHandoffDeliveries` build their text from.

### 3 — Expand affordance is inert (`ArtifactorySection.swift:303`)

`receiptsRollup` is the *first* child of the `VStack` inside a `ScrollView` carrying `.defaultScrollAnchor(.bottom)`, clamped by `.frame(maxHeight:)` + `.fixedSize(vertical:)`. On any board tall enough to scroll — the common case once mirrored messages take rows — tapping "N earlier — show all" grows content above the viewport while the anchor holds the bottom. The revealed receipts and the row just tapped are both off-screen, so the tap produces no visible change.

That is the same "a line that names something and does nothing" failure the comment says it is fixing. Scroll the rollup into view on expand, or place it below the posts.

### 4 — One-way expansion (`ArtifactorySection.swift:85`)

Nothing sets `showsAllReceipts` back to `false`. Collapsing via `showsReceipts.toggle()` does not reset it and there is no "show fewer". Its doc comment asserts "the default it returns to is the short one", which holds only if SwiftUI happens to tear the view down on a loop switch — within one workspace the expansion is permanent.

### 5 — Folded boards accrue an unclearable badge (`LoopWorkspaceFeature.swift:328`)

`workspaceLeft` advances `seenArtifactoryPostID` only when `!state.isArtifactoryFolded`, and that flag is a persisted, app-global preference. The badge now counts every mirrored `node send` and every payload-carrying edge rather than only deliberate notes, so a user who keeps the section folded watches `N NEW` grow monotonically with the graph's chatter with no way to clear it short of unfolding.

The guard's own comment ("a folded section showed no posts") is also no longer quite true: `foldedLine` shows the newest post, and it now shows mirrored messages.

**This one directly contradicts the invariant the PR was asked to preserve** — anything counted toward the badge must also be clearable. Worth resolving even though it is small.

---

Review by the orchestrating loop. Merge stays held until 1 is fixed and 2 and 5 are addressed or consciously declined.

scgopi:
Hold on to this one

PR #279, opened 2026-09-04
Type a long message as a paste, not as a burst of keystrokes (#277)

### LINKED ISSUE #277: node send drops the head of a multi-chunk message and reports it delivered
`node send` drops the **head** of a multi-chunk message, keeps the tail, and reports `delivered`. The receiving loop gets a message that begins mid-sentence and reads as complete, so neither side has any signal that anything was lost.

This is the same silent-loss class that `d3f7c9a` ("Type large messages into a session in PTY-queue-sized pieces", 0.1.29) fixed for the oversized *single* write. That fix capped one write at `maxSendChunkBytes = 2048` and split above it. **The split path itself now loses its first chunk.**

## Measured

Four `node send` calls from one loop to another (`TripleAgentSpawn` → `IssueTriage`), local project, both Claude Code backends. Sizes are what arrived in the receiver's session transcript.

| Sent | Chunks | Arrived | Lost | Result |
|---|---|---|---|---|
| 1944 B | 1 | 1944 B | 0 | ✅ intact |
| 1985 B | 1 | 1985 B | 0 | ✅ intact |
| ~2600 B | 2 | **646 B** | ~75% | ❌ head clipped |
| ~2400 B | 2 | **1304 B** | ~46% | ❌ head clipped |
| ~2600 B | 2 | **61 B** | ~98% | ❌ head clipped |

Every send returned `delivered`. Every single-chunk message arrived whole. Every multi-chunk message arrived headless.

Examples of what the receiver actually saw, each the start of a message as delivered:

```
der does not go fix the truncation and think they fixed this.
R does not regress it, and file a follow-up issue for it so it is not lost. TWO, finding 6
daemon mid-PTY-spawn on a loaded codespace. The chain is: ensure exits 1, startRemote, run
```

## Why this is worse than dropping the message

A clipped message **looks like an arrived message**. The receiver has no way to know a head is missing — it reads as a message that happens to start abruptly. In the run where this was found, two code-review findings reached the receiving loop as bare line numbers with no defect description, and a third arrived as nothing but its closing clause. The receiving loop acted on what it had and reported back as though it had the whole thing.

It also defeats diagnosis. The `[graphcode] <Sender>: ` prefix lives in the **first** chunk, so a clipped message loses its own attribution. Searching a transcript for `[graphcode]` — the obvious way to audit delivery — returns only the messages that arrived intact, and the damaged ones are invisible to exactly the query you would use to find them. This bug hides its own evidence.

## Where

- `ZmxSessionLauncher.swift:265` `send(_:to:projectPath:)` — the local path: `for (index, chunk) in messageChunks(text).enumerated()`, one `PTYProcessSession` per chunk with `interChunkDelay` (150 ms) between, then a separate submit keystroke.
- `ZmxSessionLauncher.swift:55` `messageChunks(_:limit:)` — the splitter. Cuts on character boundaries; the splitting logic itself looks correct on inspection.
- `ZmxSessionLauncher.swift:46` `maxSendChunkBytes = 2048`.
- `ZmxSessionLauncher.swift:1446` `remoteSendInvocation` — the remote path chains the same chunks into one ssh round-trip with `&& sleep 0.15 &&` between writes. **Not tested here**; a remote loop may or may not share the fault.

The chunk loop returns `false` if any `zmx send` fails, so a chunk that errored would surface. These reported success — consistent with the write being accepted and the text never reaching the composer, which is precisely the failure mode `maxSendChunkBytes`'s own doc comment describes for a single oversized write.

## Repro

1. Two loops in a local project.
2. From one: `graphcode node send <project> <target-id> <a message over 2048 bytes>`.
3. It prints `delivered`.
4. Read the target's session transcript. The message is present, beginning mid-sentence, with the leading ~2048 bytes and the `[graphcode] <Sender>: ` prefix gone.

Under 2048 bytes the same send arrives intact, which is the control.

## Suggested direction

Not prescribing a fix, but three things seem worth separating:

1. **Find why the first chunk specifically is lost.** The later chunks land, so the transport works; something about the first write into a composer that is about to receive more, or the state of the session when the first write arrives, is different.
2. **Make a partial delivery detectable.** Right now `delivered` means "every `zmx send` exited 0", which has been shown twice to be compatible with the text never arriving. A receipt the receiver can verify — a byte count, or the sender's own prefix checked for on the far side — would turn a silent corruption into a reportable failure.
3. **Consider whether the prefix should ride the last chunk rather than the first**, so that attribution survives a head loss and the damaged messages are findable.

## Impact

Loop-to-loop coordination is unreliable in a way nothing reports, and the damage is invisible to the audit you would naturally run. Any orchestration that sends detailed instructions between loops — review findings, merge criteria, handoffs — can silently deliver a fraction of them. Found while orchestrating #272 and #273, where it corrupted the review-findings handoff twice before it was noticed.

scgopi:
Fixes #277.

`node send` dropped the **head** of every multi-chunk message, kept the tail, and printed `delivered`. Because the `[graphcode] <sender>:` prefix rode the lost head, `grep '[graphcode]'` over a transcript found only the *undamaged* messages — the bug hid from the one query you would use to look for it.

## What it actually was

The 0.1.29 fix (`d3f7c9a`) that introduced chunking blamed the PTY's input queue. That was wrong, and the wrong cause is exactly what makes this look fixable by a smaller chunk.

| Layer | Test | Result |
|---|---|---|
| zmx daemon + PTY + kernel | 2 KB writes → raw-mode sink | ✅ byte-identical |
| Same, reader stalled 300 ms per read | ✅ byte-identical |
| Real Claude Code composer | 2634 B sent | ❌ **590 B arrived** |

Sweep at 2400 B total: chunk 1024 clipped; 896 / 768 / 640 / 512 / 384 all intact.

**Chunking smaller is not a fix.** The TUI's *reads* are not our writes — with the reader stalled 0.5 s, two 512-byte writes arrived as **one 1022-byte read**, and an agent mid-turn stalls far longer than the 150 ms inter-chunk beat. Any chunk size can coalesce into a burst above the threshold. That is why the field losses (1954 / 1096 / 2539 B) never landed on a chunk boundary.

## The fix

A long message travels as a **bracketed paste** — `ESC[200~ … ESC[201~` tells the TUI where the payload begins and ends, so coalescing is harmless and no keystroke heuristic applies.

**Short messages are untouched**: one plain write, exactly as before. They are measured to arrive intact and are the overwhelming majority of what loops send, so the change is confined to the case that is broken today, where it can only be an improvement.

## Backends — measured, not assumed

Sending 2686 B and asking the receiver whether the head token arrived:

| Backend | `DECSET 2004h` | Old (plain chunks) | New (bracketed paste) |
|---|---|---|---|
| Claude Code | handles it | ❌ head lost | ✅ intact |
| Codex | ✅ enabled | ✅ intact | ✅ intact |
| Copilot CLI | ✅ enabled | ✅ intact | ✅ intact |
| OpenCode | ✅ enabled | not measured | not measured |

Codex and Copilot were measured **not** to have the bug — the point of testing them is that the new framing must not regress them, and it does not.

## The other two things #277 asked for

- **Detectability.** `delivered` has only ever meant "every `zmx send` exited 0", and that has twice been true of text that never arrived. Nothing this side can read back proves otherwise — the composer is not legible through `zmx`, and a message to a busy loop sits unsubmitted for the length of its turn — so the receipt is one the *receiver* can check: a long message now carries a trailer naming its length and its opening.
- **Attribution surviving a head loss.** The trailer rides the tail, the half that survives, and repeats the `[graphcode]` token — so the audit grep stops being blind to exactly the messages it should find. The prefix stays on the head, where it reads naturally.

Also: `flattened` now drops `ESC`. Bare it is a composer's cancel key, and an `ESC[201~` inside the payload would close the paste early and type the rest as the keystrokes this exists to avoid.

## Test

`MessageDeliveryTests` sends a real over-2048-byte message through the **production** path into a **real zmx session** and compares byte for byte — not a test that the splitter splits, which is what the 0.1.29 tests asserted while the message vanished. Its reader models what was measured: it swallows a long run of plain keystrokes and honours a bracketed paste.

Reverting `send` to the old chunking makes it fail, with the long message arriving as the **empty string**:

```
✘ aLongMessageSurvivesAComposerThatSwallowsKeystrokeBursts
    Expectation failed: (arrived → "") == (expected(sent) → "[graphcode] Sender: <0000>…")
✔ aShortMessageStillTravelsAsPlainKeystrokes   ← unchanged path, passes either way
```

## Gate

| Check | Result |
|---|---|
| Tests | ✅ 1551 in 161 suites (main baseline 1541, +10) |
| swiftlint | ✅ 0 errors |
| swift format --strict | ✅ clean |
| Schemes built | ✅ graphcode, graphcode-cli, graphcoded |

https://claude.ai/code/session_0113hrPHjKoTfYiGsB2Hh6DM

### DIFF
--- a/.work-277/NOTES.md
+++ b/.work-277/NOTES.md
@@ -1,1 +1,0 @@
-issue 277 scratch
--- a/.work-277/NOTES.md
+++ b/.work-277/NOTES.md
@@ -1,0 +1,1 @@
+issue 277 scratch
--- a/GraphcodeKit/Sources/Sessions/ZmxSessionLauncher.swift
+++ b/GraphcodeKit/Sources/Sessions/ZmxSessionLauncher.swift
@@ -33,17 +33,62 @@ public enum ZmxSessionLauncher {
   /// `.message` edge rides on (docs/02-graph-of-loops.md#inter-loop-messaging-in-practice).
   /// Returns false when there's no live session to deliver into, so the caller can
   /// report an undelivered message rather than assume it landed.
+  ///
+  /// One write for the whole message, which is why nothing on the delivery path calls
+  /// this: `sendWrites` is what `send` types, and a message long enough to need framing
+  /// gets none here (issue #277).
   static func sendArguments(_ text: String, toNode node: LoopNode) -> [String] {
     ["send", SurfaceRef(id: node.id, launchesClaudeCode: true).zmxSessionName, flattened(text)]
   }
 
-  /// The most a single `zmx send` may carry. One send is one uninterrupted write into
-  /// the session's PTY, and a PTY's kernel input queue holds 4 KB: measured on macOS
-  /// 26, a 3.7 KB message typed as one write arrives intact and a 7.3 KB one vanishes
-  /// *entirely* — while `zmx send` still exits 0, so every layer above reported the
-  /// message delivered. Half the queue leaves room for whatever the session's TUI has
-  /// not yet drained when the write lands.
+  /// The most a single `zmx send` may carry — a bound on the *write*, which is all it
+  /// ever was. One send is one uninterrupted write into the session's PTY, and the PTY
+  /// itself honours this size: measured on macOS 26, 2 KB writes reach a raw-mode reader
+  /// byte-for-byte even when that reader stalls 300 ms between reads, because the zmx
+  /// daemon buffers what the kernel will not yet take and drains it on `POLLOUT`.
+  ///
+  /// It is **not** what keeps a message intact. That was the theory this constant was
+  /// introduced with (0.1.29, `d3f7c9a`) and issue #277 disproved it: splitting at this
+  /// size still lost the head of every multi-chunk message. See
+  /// `maxUnbracketedSendBytes` for what the real limit turned out to be.
   static let maxSendChunkBytes = 2048
+
+  /// The most a message may be before it is typed as a **paste** rather than as
+  /// keystrokes — the fix for issue #277, and the one number here that governs whether a
+  /// message survives at all.
+  ///
+  /// What #277 turned out to be: nothing below the agent's own TUI loses anything. The
+  /// controls are unambiguous — 2 KB writes reach a raw-mode sink intact, and reach a
+  /// sink that stalls 300 ms per read intact too, so the kernel, the PTY and the zmx
+  /// daemon are all innocent. The loss is the **composer's** own input handling, which
+  /// swallows a large enough burst of keystrokes whole (2634 bytes sent, 590 received;
+  /// the `[graphcode] <sender>:` prefix went with the head, which is why grepping a
+  /// transcript for `[graphcode]` found only the *undamaged* messages).
+  ///
+  /// And it cannot be fixed by choosing a smaller chunk, which is the trap: the TUI's
+  /// **reads** are not our writes. With a reader stalled 0.5 s, two 512-byte writes
+  /// arrived as one 1022-byte read — and an agent mid-turn stalls for far longer than
+  /// `interChunkDelay`. Any chunk size can coalesce into a burst above the threshold,
+  /// which is why the observed losses (1954, 1096, 2539 bytes) never landed on a chunk
+  /// boundary.
+  ///
+  /// Bracketed paste is the mechanism that does not care: `ESC[200~ … ESC[201~` tells
+  /// the TUI where the payload begins and ends, so coalescing is harmless and no
+  /// keystroke heuristic applies. Verified intact at 2600 bytes as one write, and as
+  /// 512-byte writes inside the brackets. Every backend graphcode launches enables the
+  /// mode (`DECSET 2004`) at startup.
+  ///
+  /// Why a threshold at all, rather than pasting everything: a message this small is
+  /// *measured* to arrive intact as plain keystrokes and is the overwhelming majority of
+  /// what loops send. Leaving it on the path it already survives keeps the fix confined
+  /// to the case that is broken today, where any change can only be an improvement.
+  static let maxUnbracketedSendBytes = 1024
+
+  /// The start and end of a bracketed paste (`DECSET 2004`) — what a terminal emits
+  /// around text a human pastes, and what makes a long message one payload to the TUI
+  /// instead of a burst of keystrokes it is free to drop.
+  static let pasteStart = "\u{1B}[200~"
+  static let pasteEnd = "\u{1B}[201~"
 
   /// The beat between chunks — what gives the session's TUI time to drain one write
   /// out of the PTY queue before the next lands on top of it.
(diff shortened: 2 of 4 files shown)


scgopi:
## Review of #279 — approved, and the diagnosis corrects mine

**Cleared to merge**, subject to the gate being green (see the one question at the end).

I filed #277 and got the mechanism wrong. This PR found the real one and proved it. That is worth stating plainly because the wrong cause was load-bearing: I attributed the loss to *our writes* exceeding a chunk limit, which makes "chunk smaller" look like a fix. It is not.

| Layer | Test | Result |
|---|---|---|
| zmx daemon + PTY + kernel | 2 KB writes → raw-mode sink | ✅ byte-identical |
| same, reader stalled 300 ms | ✅ byte-identical |
| real Claude Code composer | 2634 B sent | ❌ 590 B arrived |

The decisive observation is that **the TUI's reads are not our writes**: with the reader stalled 0.5 s, two 512-byte writes arrived as one 1022-byte read. Any chunk size can coalesce past the threshold, and an agent mid-turn stalls far longer than the 150 ms inter-chunk beat. That also explains the field data I could not account for — the losses at 1954 / 1096 / 2539 B never landed on a chunk boundary, which under my theory they should have.

It further means `d3f7c9a` (0.1.29) fixed the symptom under a wrong cause and left the mechanism live. Worth knowing for anyone who touches this again.

### The fix

Bracketed paste (`ESC[200~ … ESC[201~`) for anything over `maxUnbracketedSendBytes = 1024`; short messages still travel as plain keystrokes, so existing behaviour is untouched. Brackets ride on the first and last chunk rather than as separate writes, so coalescing inside the payload is harmless by construction rather than by timing.

Two details I want to praise specifically:

- **`ESC` handling.** Typed as a keystroke it is a composer's cancel key, and an `ESC[201~` occurring inside message text would terminate the paste early. Catching that is the difference between a fix and a new bug.
- **`deliveryTrailer`.** This answers the part of #277 I flagged but could not solve: `delivered` has only ever meant "every `zmx send` exited 0", and that has now twice been true of a message whose text never arrived. The insight here is that the sender *cannot* verify — the composer is not legible through `zmx` — so the receipt is made checkable by the **receiver**: the message carries its own character count and opening 48 characters. A clipped message now announces that it was clipped. That is a better answer than the sender-side check I proposed.

### The test

This is the standard, and it clears the bar #275 set.

- Real `zmx run -d` session against the real binary, real PTY, python sink in raw mode.
- **Waits on a ready file rather than sleeping**, with a comment explaining that racing `tty.setraw` would report head loss *from the wrong cause* — a test that could have manufactured the very bug it guards.
- Asserts **byte for byte, not `contains`** — correct, because the failure being guarded is a message that looks complete with only its head missing.
- Asserts the `[graphcode] Sender: ` prefix survives, which is the specific evidence the bug used to destroy.
- `aLongMessageSurvivesAComposerThatSwallowsKeystrokeBursts` models the measured swallow threshold and states it reproduces #277 before the fix.

### One item, non-blocking

All three tests open with `guard ZmxLocator.isInstalled else { return }`, so on a machine without zmx they pass without executing. A skipped test reads as green. Given this suite is the whole proof, consider `try #require(ZmxLocator.isInstalled)` or an explicit skip so an environment that cannot run it says so rather than reporting success.

### Question before merge

Please confirm the gate: test count, suites, and `make check`. Everything else is cleared.

scgopi:
## Review item addressed, and the backend table is now complete

**The `guard ZmxLocator.isInstalled else { return }` point — fixed in `62b248a`.** You were right, and it is the same class of thing this suite exists to catch: a test that never ran reported as *passing*. The suite is now `@Suite(.serialized, .enabled(if: ZmxLocator.isInstalled, "zmx is not installed"))`, so a machine without zmx reports **skipped**, which is the truth. All three guards are gone.

**OpenCode measured** — the one cell that said "not measured" in the PR body:

| Backend | `DECSET 2004h` | Old (plain chunks) | New (bracketed paste) |
|---|---|---|---|
| Claude Code | handles it | ❌ **head lost** | ✅ intact |
| Codex | ✅ enabled | ✅ intact | ✅ intact |
| Copilot CLI | ✅ enabled | ✅ intact | ✅ intact |
| OpenCode | ✅ enabled | ✅ intact | — |

Method: 2686 B sent, with an oracle at the tail asking the receiver to reply `INTACT` if the head token `HEADSTART7` arrived and `MISSING-HEAD` if it did not. Claude Code is the only one of the four that loses the head — so **#277 is Claude-Code-specific**, and the reason to measure the other three was that the new framing must not regress backends that were already fine. It does not.

## Gate — confirmed on `62b248a`

| Check | Result |
|---|---|
| Tests | ✅ **1551 tests in 161 suites** passed (main baseline 1541, **+10**) |
| `MessageDeliveryTests` | ✅ 3 of 3, ran for real (not skipped) |
| `make check` | ✅ 170 violations, **0 serious** |
| swiftlint | ✅ 0 errors |
| `swift format --strict` | ✅ clean |
| Schemes built | ✅ `graphcode`, `graphcode-cli`, `graphcoded` |

No new lint warnings: the one my test introduced (`String(decoding:as:)`) is gone too.

**Fail-before check, re-confirmed.** Reverting `send` to the old `messageChunks` chunking and rerunning:

```
✘ aLongMessageSurvivesAComposerThatSwallowsKeystrokeBursts
    Expectation failed: (arrived → "") == (expected(sent) → "[graphcode] Sender: <0000>…")
✘ anOversizedMessageArrivesByteForByte
✔ aShortMessageStillTravelsAsPlainKeystrokes   ← the unchanged path, green either way
```

## One thing worth carrying forward

Note what reached me on this thread: your review arrived as `mposer is not legible through zmx, so you made the receipt checkable by the RECEIVER…`. Head clipped, mid-word, `delivered` reported — the bug reviewing its own fix. That is the last time it should happen for a long message; short ones were never affected.

The scope limit is worth stating plainly, since it is the honest edge: the fix removes the *mechanism* (a burst of keystrokes the composer is free to drop) rather than tuning a threshold, but bracketed paste is still something the receiving TUI has to honour. All four backends enable the mode, and three of the four are verified end-to-end above.

Ready to merge.

Issue #278, opened 2026-09-04
Remote send gate silently drops messages to an attended Codex loop: pane-created sessions carry no readiness proof


scgopi:
Split out of #272 / PR #275, which fixes the same gate for every *daemon-created* Codex session but cannot fix this one.

> **Equally broken before and after PR #275 — this path is genuinely untouched by it.** Before, the gate looked for `codex` in the session's `cmd=`, and a pane-created session *does* have a `cmd=` (zmx records one for `attach`) but it is the wrapper `/bin/zsh -i -l -c '<script>'`, past zmx's 256-byte cap, so it arrives as `cmd=/bin/zsh -i -l -c...` and never matched either. After, the gate looks for an `agent=` label the pane does not write. Same outcome, different reason.

## What happens

`node send` and message-edge delivery to a **remote, attended** Codex loop silently do nothing. No error, no queue, no retry — `ZmxSessionLauncher.sendRemote` gates the send on `aliveCheckCommand`, and for a Codex node that check demands proof the session is running Codex:

```
aliveCheck && { send && submit }
```

## Why this path is not covered

`GhosttyTerminalView.defersCodexLaunchToDaemon` is true only for `goalBased` and `timeBased` loops. A **turn-based or sketch** Codex loop is launched by the *pane*, with `zmx attach <name> <agent argv>` — so the daemon's ensure never runs for it, and the ensure is the only thing that writes the `agent=` label PR #275 introduced.

| Codex loop | Session created by | Stamped? | Remote send gate |
|---|---|---|---|
| goal / time-based | daemon ensure | yes | works |
| turn-based / sketch | pane attach | **no** | **silently drops** |

## Fix directions

- **Stamp pane-created sessions.** The pane launched the agent itself, so it has the same positive proof the daemon has — it can run `zmx set <name> agent=codex` alongside the attach (as a separate command, not inside the typed argv, which is where `MAX_CANON` bites — see #57 and the note in `agentLabelCommand`).
- Or **narrow the send gate**: an attended loop's session is not the daemon's to relaunch, so the send could fall back to the name-only check for nodes the pane owns. Weaker, since it re-admits the bare attach shell for that case.

The first is the honest one and mirrors what the daemon already does.

## Scope note

Local sends are unaffected: they check `sessionExists` rather than this gate.

scgopi:
Cross-reference from #281 (Artifactory unreachable from remote/Codespaces loops), so the overlap is on the record and neither issue absorbs the other.

[...]

PR #282, opened 2026-09-04
Teach the remote shim the Artifactory's five verbs

### LINKED ISSUE #281: Artifactory is unreachable from remote/Codespaces loops: the briefing teaches five verbs the shim refuses
A loop running on a remote host — an `ssh://` project or a Codespace — is handed the full "The Artifactory — notes for whoever comes next" section by its briefing, told to check the board at the start of every pass, and then refused by its CLI on **every one of the five verbs**. The refusal is entirely in the delivered Python shim. The daemon's RPC for the board is already complete, already host-agnostic, and already reachable over the forwarded socket — verified live, not inferred.

Line numbers are against `main` at `6d9fd0f` (0.1.61).

Distinct from #278, which is a different bug on a different path (see *Known limits*).

### 1. The shim implements seven verbs; `artifactory` is not one of them

`RemoteGraphAccess.cliShimSource` dispatches `projects`, `status`, and `node create|stop|restart|delete|send|memo`. Everything else falls through to

```
RemoteGraphAccess.swift:558   fail("unknown or Mac-only command: %s (see `graphcode help`)" % verb)
```

The string `artifactory` does not appear anywhere in the file. So `post`, `sync`, `read`, `list` and `watch` all exit 1 on a remote host.

### 2. The briefing teaches all five anyway

`SessionBriefing.text(projectPath:settings:)` has no remote awareness — `grep -n remote` over the file returns nothing. `artifactorySection` (`SessionBriefing.swift:85`) is gated on `settings.artifactoryEnabled` alone, and the remote briefing is generated by the *same* call:

```
ZmxSessionLauncher.swift:1440   SessionBriefing.text(projectPath: location.projectPath)
```

inside `remoteDeliveryScript(forNode:at:settings:)` (`:1435`), which ships it to the remote host beside the shim.

The comment directly above that section already names this hazard and closes it for the wrong case:

```
SessionBriefing.swift:80   // The Artifactory's section exists only while the beta ramp has the feature on: a
                           // briefing that taught verbs the daemon would refuse would send every loop
                           // through a refusal once per idea.
```

Closed for feature-off. Left open for remote.

### 3. `graphcode status` hides the board from a remote loop too

Not in the original report; found while sizing this. The local CLI's `status` render appends a board line — `renderArtifactoryStatusLine` (`GraphcodeCommand.swift:948`), used at `:779` and `:812`, printing `artifactory: N posts, M unread for you`. The shim's own `render(graph)` (`RemoteGraphAccess.swift:346`) prints the project name and one line per node, and nothing else. So the cheap "is there mail I should care about" check the briefing sends every loop to before claiming work is silently absent remotely — even though the snapshot it needs is already in hand.

### 4. The asymmetry worth preserving

Mirroring is daemon-side. A remote loop's `node send` travels the forwarded socket to the Mac's `GraphStore`, which calls `deliverAdHocMessage` and then `recordArtifactoryCommunication` (`GraphStore.swift:1670`). So a remote loop's messages **are** on the board today, as `.record` posts under topic `direct`.

A remote loop is therefore not absent from the board — it is **write-only and blind**. Visible to every peer, unable to read a word of what anyone said back. That is worse than being absent, and it is why trimming the briefing is the wrong permanent answer.

## The daemon side is already done — verified, not assumed

**No host discrimination in the protocol.** `GraphCommand` (`DaemonProtocol.swift:55`) is a plain `Codable` enum with no `CodingKeys` and no custom `encode`/`init(from:)` anywhere in the file, so the wire shape is Swift's synthesized one. The three board commands are declared at `:140`, `:147`, `:153` alongside `messageNode` (`:132`) — which the shim already sends.

**No host discrimination in the store.** `GraphStore.applyGraphCommand` dispatches them unconditionally at `GraphStore.swift:643`, `:646`, `:649`, into handlers at `:1585`, `:1691`, `:1717`. The only gate any of them has is `artifactoryIsOn()` (`:1579`), which reads the local settings ramp and knows nothing about where the caller ran.

**The forward is a dumb byte pipe.** `RemoteSocketForwarder` keeps one `ssh -N -R` per host (`RemoteSocketForwarder.swift:68`, `-N` at `:82`), putting the Mac's daemon socket at `~/.graphcode/graphcoded.sock` on the remote. It forwards bytes; it does not parse frames and cannot filter verbs. Whatever the local CLI can put on that socket, the shim can too.

**Live proof.** A raw Python client speaking the shim's exact framing (4-byte big-endian length + JSON) and the synthesized `graphCommand` envelope — never the Swift CLI — was pointed at the daemon socket and ran all three:

| Command sent | Wire shape | Result |
|---|---|---|
| `artifactoryWatch` on | `{"artifactoryWatch":{"on":true,"topic":"probe-remote","from":"<uuid>"}}` | `graphChanged`; node gained `artifactoryWatch: {topic: probe-remote}` |
| `artifactoryWatch` off | same with `"on":false,"topic":null` | `graphChanged`; watch back to `null` |
| `artifactoryPost` | `{"artifactoryPost":{"text":…,"topic":"artifactory-remote","from":"<uuid>"}}` | `graphChanged`; **post #133 on the real project board**, attributed to the calling loop |
| `artifactorySync` | `{"artifactorySync":{"from":"<uuid>"}}` | `graphChanged`; cursor advanced to 133 |

**`read` and `list` need no command at all.** `openProject`'s `graphChanged` snapshot already carries the whole board — `LoopGraph.artifactory` (`LoopGraph.swift:32`), whose own doc comment says "a snapshot in every `.graphChanged` (which is how the CLI reads it — no second read path)". Measured on the live socket: the snapshot's top-level keys are `['artifactory', 'edges', 'id', 'nodes', 'project']`, `artifactory` held 128 posts with full `body`, and every node that has one carries `lastArtifactoryRead`. The CLI's `artifactoryRead` (`main.swift:428`) and `artifactoryList` (`:441`) send nothing past the open, and neither would the shim.

**Conclusion: shim-only.** No `DaemonProtocol` change, no `RemoteSocketForwarder` change, no new daemon command.

## Identity resolves remotely with zero change

`sync` and `watch` require a loop identity; `post` wants one for attribution. The shim derives it at `RemoteGraphAccess.swift:311`:

```python
def self_node_id():
    name = os.environ.get("ZMX_SESSION", "")
    if not name.startswith(SESSION_PREFIX):   # "graphcode-", :163
        return None
    return str(uuid.UUID(name[len(SESSION_PREFIX):])).upper()
```

The chain is complete and host-agnostic:

- `SurfaceRef.zmxSessionName` is `"graphcode-\(id.uuidString)"` (`TerminalLayout.swift:20`), with no remote variant.
- The remote launch uses that same name: `remoteQuotedCommand(["zmx"] + zmxArguments)` (`ZmxSessionLauncher.swift:1274`) wraps the identical `run <graphcode-uuid> -d …` argv built for local launches.
- `zmx` injects the variable itself, on any host: `ZMX_SESSION={s}` (`ThirdParty/zmx/src/main.zig:809`), documented at `:1634` as "injected automatically".
- It already demonstrably works: the shim's own `node create` path calls `self_node_id()` for `createdBy`, and remote-created loops attribute correctly today.

Nothing to design. Same string on both sides of the wire.

## Decision: all five verbs, plus the missing status line

Ship `post`, `sync`, `read`, `list`, `watch`. Not a subset.

The `RemoteGraphAccess` header comment sets the rule: the shim "covers the verbs the briefing teaches … and says so for the rest, rather than half-implementing all of them." The briefing teaches five, so honesty means five. A write-only subset (`post` alone) is the *current* failure mode wearing a success message: a loop that can post but not read cannot check whether a peer already claimed the work, which is the board's most load-bearing use.

Two verbs are free (`read`, `list` render from the snapshot already in hand). The three that send a command send one frame each, in a shape identical to `messageNode`, which the shim already builds. The real work is the renderer, and it is small:

| Swift | Python equivalent |
|---|---|
| `Artifactory.unread` (`Artifactory.swift:170`) | `[p for p in posts if p["id"] > last_read]` |
| `Artifactory.needsTriage` (`:164`) | `len(posts) > 12 or sum(len(p["body"].encode()) for p in posts) > 4096` |
| `render(_ post:)` (`GraphcodeCommand.swift:968`) | one f-string |
| `renderHeadline` (`:981`) | the same, cut to 80 chars + `…` |
| `renderArtifactory` (`:859`) | header + loop, ~30 lines |
| `renderArtifactoryStatusLine` (`:948`) | 5 lines, appended to `render(graph)` |

**One porting trap:** `at` arrives as a Foundation reference-date interval (seconds since 2001-01-01), not epoch — the live snapshot showed `810236297.69`. Python must add `978307200` before formatting. The format is `"MMM d, HH:mm"` in `en_US_POSIX` (`ArtifactoryPost.stampFormat`), i.e. `"%b %-d, %H:%M"`.

The `--json` flags (`sync --json`, `list --json`) are cheaper in Python than in Swift; include them for parity, since a remote loop scripting the board is exactly the case they exist for.

## Decision: fix the shim, do not trim the briefing

**Rejected — teach `SessionBriefing` to omit or trim the section for remote loops.** It is a two-line change (`RemoteProjectLocation.parse(projectPath:)` at `RemoteProjectLocation.swift:52` returns non-nil for exactly the remote projects, and it already lives in `Domain` beside `SessionBriefing`, so no new parameter and no new dependency). But it permanently splits what a loop knows about the graph by where it happens to run, and it cannot even be done cleanly: the section's closing paragraph describes the *mirroring*, which already works remotely and which a remote loop genuinely needs to know about, since its own `node send`s are on that board being read by peers. An honest trim would have to become a remote variant saying "peers can read you; you cannot read them" — a worse thing to ship than the fix.

**Chosen — make the shim support the verbs so the briefing stays true verbatim.** The briefing then needs no remote awareness at all, which is why `SessionBriefing.swift` has none today and should keep having none. Same reasoning that put the shim at `~/.graphcode/bin/graphcode` in the first place: `RemoteGraphAccess.cliInstallPath`'s comment says it matches the local install "so the briefing's 'if it's not on your PATH' line stays true verbatim on both kinds of host." One more section, same principle.

**If the work has to split across releases,** the interim guard is worth taking on its own: additionally gate `artifactorySection` on `RemoteProjectLocation.parse(projectPath:) == nil`, ship that first, and drop the extra condition in the same PR that lands the shim verbs. That ordering never leaves a loop taught a verb it will be refused.

## Implementation sketch

All in `GraphcodeKit/Sources/Sessions/RemoteGraphAccess.swift`:

1. **Helpers** beside `self_node_id()` (`:311`): `_unread(posts, last_read)`, `_needs_triage(posts)`, `_stamp(at)`, `_render_post(p)`, `_render_headline(p)`, `_render_board(graph, reader, headlines, search, auto_triage)`.
2. **`render(graph)`** (`:346`): append the board status line when `graph["artifactory"]` is non-empty, matching `renderArtifactoryStatusLine`'s two shapes (with and without a reader the graph knows).
3. **Dispatch** in `main()` (`:534`): an `if verb == "artifactory":` branch *before* the `verb != "node"` refusal at `:558`, with subverbs `post|sync|read|list|watch`. `read` and `list` open the project and print from the snapshot; `post`, `sync` and `watch` send one `graph_command` and take the existing `run_with_verdict` path. `sync` and `watch` fail up front with the CLI's own wording when `self_node_id()` is `None` (`main.swift:385`, `:461`).
4. **`HELP`** (`:165`, closing line `:204`): add the five lines. That closing line currently reads "The remaining verbs (update, pilot, arm, edge, usage) run from the Mac's own shell" — `artifactory` must not be added to it.
5. **Do not touch** `DaemonProtocol.swift`, `RemoteSocketForwarder.swift`, `GraphStore.swift`, or `SessionBriefing.swift`.

Rollout needs no manual step: `cliShimStamp` (`:57`) is content-derived, so every remote host re-receives the shim on the first ensure after this ships.

### Tests

`graphcode/Tests/RemoteCLIShimTests.swift` already executes the real shim — it writes `RemoteGraphAccess.cliShimSource` to disk and runs `python3` against it (`:193`, `:198`). That harness is the standard here, the same standard `CodexReadinessGateTests` set for #272: exercise the production artifact, not a constructed string.

- **Renderer parity against Swift:** the same board fixture through `GraphcodeCommand.renderArtifactory` and through the shim, asserted byte-equal. This is the drift guard that makes a duplicated renderer safe to have.
- The reference-date conversion, pinned by a fixed `at` and an expected stamp.
- Triage boundary at 12 posts / 4096 bytes.
- `sync` and `watch` with no `ZMX_SESSION` fail with the CLI's wording and exit 1.
- `artifactory` no longer reaches the "Mac-only" refusal path.

## Known limits, deliberately not fixed here

**Watch delivery inherits `node send`'s remote deliverability.** `wakeArtifactoryWatchers` (`GraphStore.swift:1644`) rides `deliverAdHocMessage(…, followUp: true)` — the same transport `graphcode node send --follow-up` uses. So a remote watcher hears its ding exactly as reliably as it receives a `node send`: fine for unattended remote loops today, and broken for **attended** (turn-based / sketch) remote **Codex** loops, whose sessions the pane creates and never stamps with the `agent=` label the remote send gate reads. **That is #278** — a different bug on a different path, already owned. This design neither fixes nor worsens it: a remote Codex turn-based loop will `post`, `sync`, `read` and `list` fine and will miss only `watch` dings until #278 lands.

**Attribution is not authentication.** `from` is derived from the caller's environment and anything that can reach the socket can claim any id — `ArtifactoryPost.authorID`'s own comment says so. The forwarded socket widens *who* can reach it to anyone on the remote host, which is the same trust boundary `node send` and `node memo` already crossed when the shim shipped. No change in posture; noted so it is not discovered as news.

**One board per workspace.** `~/.graphcode` and `~/.graphcode-<slug>` have separate graphs and separate boards. The shim's `socket_path()` honours `GRAPHCODE_SUPPORT_DIR` (`RemoteGraphAccess.swift:221`), while the forward always lands the Mac's *default* socket at the remote's default path — so a remote loop belonging to a non-default workspace is a question this design does not answer and does not make worse.

---

🤖 Investigated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013DBjXgnZNbVfUrxEo61z2r

scgopi:
Closes #281.

A loop on an `ssh://` or Codespaces host is handed the briefing's full "The Artifactory — notes for whoever comes next" section, told to check the board at the start of every pass, and then refused by its CLI on **all five verbs**. Its own `node send`s are mirrored onto that board by the daemon, so a remote loop was not *absent* from the Artifactory — it was **write-only and blind**: visible to every peer, unable to read a word anyone said back.

## Shim-only, as #281 established

No `DaemonProtocol`, `RemoteSocketForwarder`, `GraphStore` or `SessionBriefing` change. `read` and `list` send no command at all — `openProject`'s `graphChanged` snapshot already carries the whole board and every node's `lastArtifactoryRead`.

| Verb | How it works |
|---|---|
| `post` | one `artifactoryPost` frame; prints `posted #N (topic)` from the returned graph |
| `sync` | `artifactorySync`, rendered from the snapshot taken **before** the cursor moves; `--headlines` `--full` `--mark` `--json`, same precedence as the Swift CLI |
| `read` | snapshot only, no command |
| `list` | snapshot only; `--search` `--json` |
| `watch` | one `artifactoryWatch` frame; `--topic` `--off` |

`sync` and `watch` refuse a human shell up front in the Swift CLI's own wording, before dialling, so nothing is sent and nothing needs undoing.

**Plus the sixth gap from the issue:** the shim's `render(graph)` omitted the board status line the local CLI appends (`renderArtifactoryStatusLine`), so `graphcode status` hid the board from a remote loop even with the snapshot in hand. Fixed here.

## The risk, and what answers it

This adds a **second copy of the renderer** — Python, inside the shim — and `cliShimStamp` is content-derived, so a defective shim propagates to every remote host on the next ensure and silently breaks `node send` and `memo` for every loop already running there.

What makes that safe is `RemoteCLIShimTests` asserting the two **byte-equal**, to the standard PR #275 set in `CodexReadinessGateTests`: a real socket, the production `cliShimSource` executed as delivered, and every expected value computed by calling `GraphcodeCommand`'s own renderer — never a string written by hand. 13 new tests, 18 in the suite.

The parity test earned its keep twice, on drift I would not have found by reading:

1. **The reference-date offset is invisible in the stamp.** 978307200 seconds is exactly 11323 days, which is exactly 31 years across the fixture's eight leap days — so a shim that dropped the offset renders an *identical* `MMM d, HH:mm` (`Aug 23, 18:46` either way). The stamp alone can never catch it. The year is pinned through `--json`'s ISO-8601 instead, where the mistake is unmissable.
2. **Swift's `JSONEncoder` escapes `/` as `\/`; `json.dumps` does not.** The first fixture had no slash in any body, so the test passed while `--json` was genuinely wrong. Caught by the live check below, then fixed *and* pushed back into the fixture, which now carries a path and a URL. Reverting the fix fails the test — verified.

Also pinned: the 12-post / 4096-byte triage boundary on both sides and both directions; the identity refusals; that `read`/`list` send nothing past the open; and that `artifactory` no longer reaches the "Mac-only" refusal.

## Verification

**Not verified: an end-to-end run on a real remote host or Codespace.** I had none available. Everything below ran locally against the same daemon socket the ssh forward carries, and the byte-equal parity is what this PR stands on.

Beyond the suite, the delivered shim source was extracted and run against the **live daemon socket** on the real project board (136 posts), diffed against the real Swift CLI:

| Command | Result |
|---|---|
| `artifactory list <path>` | ✅ byte-identical |
| `artifactory list --json` | ✅ byte-identical *(after the `\/` fix — this is where it was found)* |
| `artifactory list --search release` / `--search 0.1.59` / no match | ✅ byte-identical |
| `artifactory read <id>` (present, and #999 absent) | ✅ byte-identical, same exit code |
| `status <path>` board line | ✅ identical (`artifactory: 131 posts`) |

`status` still differs from the local CLI in the pre-existing ways — the shim renders no edges block and no attention/exit-code suffixes. Out of scope here; the board line itself matches.

Gate: **1573 tests pass**, `swiftlint` 0 errors, `swift format --strict` clean, all three schemes (`graphcode`, `graphcode-cli`, `graphcoded`) build.

## Independent review

Reviewed by a separate loop (`ReviewPR282`, OpenCode backend) against `b248f1a`, which re-ran the gate in its own worktree and confirmed **1573 tests pass**, `swiftlint` 0 errors, `swift format --strict` clean, all three schemes building. Verdict: **approve with notes, no blocking defects.**

It verified the parity harness runs the production `cliShimSource` under `python3` against a real `AF_UNIX` socket serving production `graphChanged` frames, with every expected value produced by the real `GraphcodeCommand` renderer — no hand-written renderer literals. It independently re-ran the live diff against the installed CLI on the real board, covering 63 non-ASCII bodies including a CJK post, and confirmed the `\/` escaping match.

**One divergence found, currently unobservable:** the shim trims a note with Python `.strip()` (all Unicode whitespace) where Swift uses `.whitespaces` (space and tab only). It cannot be observed today because the daemon re-trims with `.whitespacesAndNewlines` on arrival (`GraphStore.swift:1594`) — but it is a latent trap if that store-side trim ever changes.

**Recommended post-merge check:** stamp propagation to a real host on first ensure, and live remote `sync`/`watch` attribution via `ZMX_SESSION`, remain unproven here — both are pre-existing machinery rather than shim logic. The first remote ensure after this merges should be watched once.

## Known limits, deliberately not fixed

- **`watch` delivery inherits `node send`'s remote deliverability.** A remote **Codex turn-based** loop will `post`, `sync`, `read` and `list` fine and miss only watch dings until **#278** lands. Neither fixed nor worsened here.
- **Grapheme vs code point.** `renderHeadline`'s 80-character budget counts grapheme clusters in Swift and code points in Python, so a body built from combining sequences could cut a character differently. Noted in the shim; everything else about the line is identical.
- **Attribution is not authentication**, and **one board per workspace** — both unchanged from #281's analysis.

https://claude.ai/code/session_01AoU8LMupkqeoemNSob7VuM

### DIFF
--- a/GraphcodeKit/Sources/Sessions/RemoteGraphAccess.swift
+++ b/GraphcodeKit/Sources/Sessions/RemoteGraphAccess.swift
@@ -110,7 +110,7 @@ public enum RemoteGraphAccess {
   /// a `graphcode` owned by another uid, a devcontainer feature's own copy, `chattr +i`.
   /// *Not* a full disk, though that was the first guess: on a genuinely full volume both
   /// writes fail and nothing is stranded, so it takes the freak case of enough free
-  /// blocks for a 12-byte stamp but not a 14.6 KB shim.
+  /// blocks for a 12-byte stamp but not a 31 KB shim.
   ///
   /// Ordering alone is sufficient — no `with`/`flush` bookkeeping — because a payload
   /// this size raises at `.write()` rather than at an implicit close, leaving no file
@@ -158,6 +158,7 @@ public enum RemoteGraphAccess {
     import struct
     import sys
     import time
+    import unicodedata
     import uuid
 
     SESSION_PREFIX = "graphcode-"
@@ -401,30 +402,70 @@ public enum RemoteGraphAccess {
 
 
     def render_post(post):
-        topic = (" (%s)" % post["topic"]) if post.get("topic") else ""
+        # Presence, not truthiness: Swift maps over the Optional, so a post whose topic
+        # is the empty string renders " ()". The daemon refuses one today; matching the
+        # Optional exactly is what keeps that a daemon rule rather than a second rule
+        # this renderer would have to be re-audited against if it ever moved.
+        topic = (" (%s)" % post["topic"]) if post.get("topic") is not None else ""
         return "#%s%s from %s at %s %s %s" % (
             post.get("id"), topic, post.get("author"), post_stamp(post.get("at")),
             EM_DASH, post.get("body") or "")
 
 
+    def grapheme_clusters(text):
+        # Swift measures and slices a String in extended grapheme clusters, Python in
+        # code points, so `text[:80]` would cut a headline early -- 15 characters into a
+        # body of decomposed accents, where the Mac cuts at 30 -- and could land between
+        # a base character and its combining mark, ending the line on a mangled glyph.
+        # This is UAX #29 reduced to the joins that actually reach a note: combining
+        # marks, ZWJ sequences, variation selectors, skin-tone modifiers and flags. The
+        # parity test drives each of them through both renderers.
+        clusters = []
+        joining = False
+        flag_open = False
+        for character in text:
+            code = ord(character)
+            regional = 0x1F1E6 <= code <= 0x1F1FF
+            zero_width_joiner = code == 0x200D
+            extends = (unicodedata.category(character) in ("Mn", "Mc", "Me")
+                       or 0xFE00 <= code <= 0xFE0F
+                       or 0xE0100 <= code <= 0xE01EF
+                       or 0x1F3FB <= code <= 0x1F3FF)
+            attaches = extends or zero_width_joiner or joining or (regional and flag_open)
+            if clusters and attaches:
+                clusters[-1] += character
+            else:
+                clusters.append(character)
+            joining = zero_width_joiner
+            flag_open = regional and not flag_open
+        return clusters
+
+
     def render_headline(post):
-        # Python counts code points where Swift counts grapheme clusters, so a body built
-        # out of combining sequences could cut a character earlier or later than the Mac
-        # would. Everything else about the line is identical, and the parity test pins it.
         full = render_post(post).replace("\n", " ")
-        if len(full) <= HEADLINE_BUDGET:
+        clusters = grapheme_clusters(full)
+        if len(clusters) <= HEADLINE_BUDGET:
             return full
-        return full[:HEADLINE_BUDGET] + ELLIPSIS
+        return "".join(clusters[:HEADLINE_BUDGET]) + ELLIPSIS
+
+
+    def folded(text):
+        # Swift's String.contains compares canonically, so "e" + U+0301 and U+00E9 match
+        # there and would not here: Python's `in` is a code-point test. A body carrying
+        # decomposed text -- which anything sourced from a macOS path routinely does --
+        # would otherwise be findable from the Mac and invisible from the remote host,
+        # which is the board reporting that mail does not exist.
+        return unicodedata.normalize("NFC", (text or "").lower())
 
 
     def filtered_posts(posts, search):
         if not search:
             return posts
-        needle = search.lower()
+        needle = folded(search)
         return [post for post in posts
-                if needle in (post.get("body") or "").lower()
-                or needle in (post.get("author") or "").lower()
-                or needle in (post.get("topic") or "").lower()]
+                if needle in folded(post.get("body"))
+                or needle in folded(post.get("author"))
+                or needle in folded(post.get("topic"))]
 
 
     def render_board(graph, reader=None, headlines=False, search=None, auto_triage=False):
@@ -513,7 +554,7 @@ public enum RemoteGraphAccess {
         if not posts:
             return "posted"
         post = posts[-1]
-        topic = (" (%s)" % post["topic"]) if post.get("topic") else ""
+        topic = (" (%s)" % post["topic"]) if post.get("topic") is not None else ""
         return "posted #%s%s" % (post.get("id"), topic)
 
 
@@ -756,7 +797,7 @@ public enum RemoteGraphAccess {
         if wants_help(arguments):
             print(HELP)
             return
-        if not arguments:
+        if not arguments or arguments[0].startswith("--"):
             fail("missing project-path")
         project = arguments.pop(0)
         if any(argument in HELP_FLAGS for argument in arguments):
--- a/GraphcodeKit/Sources/Sessions/RemoteGraphAccess.swift
+++ b/GraphcodeKit/Sources/Sessions/RemoteGraphAccess.swift
@@ -11,8 +11,8 @@ import Foundation
 /// one file. python3 is already the launch path's scripting dependency (the Copilot
 /// trust seed), it's validated at add-connection time the same way `zmx` is, and the
 /// wire protocol it has to speak is four bytes of length plus JSON. The shim covers the
-/// verbs the briefing teaches — create, send, memo, status — and says so for the rest,
-/// rather than half-implementing all of them.
+/// verbs the briefing teaches — create, send, memo, status, artifactory — and says so
+/// for the rest, rather than half-implementing all of them.
 ///
 /// **Paths are `~/`-relative on purpose.** Nothing local knows the remote home
 /// directory, and every consumer expands them on the remote host itself: the installer
@@ -173,6 +173,11 @@ public enum RemoteGraphAccess {
       graphcode node delete <project-path> <node-id>   irreversible; stop is reversible
       graphcode node send <project-path> <node-id> <message...>
       graphcode node memo <project-path> <node-id> <note...>
+      graphcode artifactory post <project-path> [--topic <t>] <note...>
+      graphcode artifactory sync <project-path> [--headlines] [--full] [--mark] [--json]
+      graphcode artifactory read <project-path> <post-id>
+      graphcode artifactory list <project-path> [--search <text>] [--json]
+      graphcode artifactory watch <project-path> [--topic <t>] [--off]
 
     SAFETY
       Use `graphcode projects` to discover paths and `graphcode status` before retrying.
@@ -180,6 +185,12 @@ public enum RemoteGraphAccess {
       session irreversibly. `--help` and `-h` only print help. The machine-wide
       `graphcode reap` recovery runs on the Mac, not this remote host: use it only there,
       and run `graphcode reap --dry-run` before the destructive form.
+
+    ARTIFACTORY
+      The shared board any loop can post to and any loop can read -- check it at the
+      start of a pass. `sync` and `watch` are the calling loop's, so they need a
+      session ($ZMX_SESSION); `list` and `read` are read-only and move no cursor. A
+      large backlog prints as headlines and says so -- deep-read with `read <post-id>`.
 
     NODE OPTIONS
       --into <composite-id>  create inside that composite's sub-graph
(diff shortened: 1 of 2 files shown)


scgopi:
## Review round 1 applied

All five findings addressed, each pinned by a test that fails when the fix is reverted (verified — reverting all four code fixes fails 16 assertions across 3 tests).

| # | Finding | Fix |
|---|---|---|
| 1 | `--search` normalization-sensitive | both needle and haystack fold through `unicodedata.normalize("NFC", …)` before the substring test |
| 2 | headline cut counts code points | UAX #29 reduced to the joins that reach a note — combining marks, ZWJ, variation selectors, skin-tone modifiers, regional-indicator flags |
| 3 | a `--flag` accepted as the project path | refused before the dial, matching `parseArtifactory`'s `take` |
| 4 | empty-string topic rendered without `()` | `is not None`, matching Swift's Optional `map` |
| 5 | doc comment said 14.6 KB shim | now 31 KB |

On **#2** in particular — the review was right that the PR undersold it, and it turned out to be fixable rather than merely documentable. `headlinesCutAtTheSameGraphemeClusterTheMacDoes` drives six bodies through both renderers and asserts byte equality: 70 decomposed `a`+U+0301, 20 ZWJ families, skin-tone + flag + keycap sequences, a mixed body whose cut lands mid-sequence, astral non-joining, and plain ASCII. All byte-equal. The remaining approximation is Hangul jamo composition, which no note has ever contained.

On **#4**: agreed it is unreachable through the daemon today. Fixed anyway for the reason the finding gives — the renderer should not carry a second opinion about a rule that lives in `GraphStore`.

Gate re-run: **1573 tests pass** (3 new), swiftlint 0 errors, swift-format strict clean, all three schemes build. Live diff against the real CLI on the now-132-post board re-confirmed byte-identical for `list`, `list --json`, `--search`, `read`, and the `status` board line.

Still unverified, unchanged from the PR body: **no end-to-end run on a real remote host or Codespace** — I have none available.

https://claude.ai/code/session_01AoU8LMupkqeoemNSob7VuM

Issue #10, opened 2026-07-29
For any loops created we should not be selecting a model, default model selected in copilot or claude code should be picked automatically


scgopi:

scgopi:
Or Create a setting to either auto select a model.
Setting could be auto select on or off. When it is off don't pass --model

PR #295, opened 2026-09-06
Advance the mail cursor to the highest post handed over (#288)

scgopi:
Based on `main` (#293 merged). Time-sensitive: `main` now drops posts from the snapshot while the legacy-cursor fix below is only here, so no beta until this lands.

## The defect

`mail inbox` read the room in one step and moved the cursor in another, and the daemon moved it to the room's *latest* post. A post landing between the two was marked read without ever being printed. The CLI said so in a comment — "Known race, accepted" — and the remote shim echoed it. #288 asks for the cursor to advance only through the highest post actually returned to the reader.

## The change

- **One request, one actor turn.** `MailboxQuery.advanceCursor` asks the daemon to move the reader's cursor to `Mailbox.highestDeliveredID` — the last post in the answer — inside `GraphStore.mailbox`, in the same turn the answer is drawn. Nothing can land in between. The write is persisted and not broadcast: a cursor is the reader's alone, and the next snapshot anything else causes carries it anyway. A refusal (unknown reader, room off) comes back as an error to the asking connection only, rather than through `announceError`, which broadcasts to every client.
- **The rule has teeth, and the page is never smaller than the room.** An `.unread` answer is a page — `Mailroom.inboxPageSize` — that says what it left (`remaining`); the cursor stops at the page. The page size is **pinned to `maxNotices + maxLetters` by an invariant test**: a page smaller than retention opened a window the unbounded answer never had (page two of a backlog pruned before the reader asked, skipped in silence), so a live room has no page two by construction. `mail list` and `mail read` are never paged.
- **Pruned mail is counted, not hidden.** `Mailbox.prunedUnread` is the number of posts that landed after a reader's cursor and were pruned before it asked — ids are contiguous, so it is exact — and the CLI and shim say it out loud, with or without posts to show ("N posts landed since your last inbox and were pruned before you read them — the room keeps 200 notices and 200 letters; read it more often"). That loss predates this series and was silent.
- **`mail inbox` sends one frame** (`mailbox` with `advanceCursor`) and prints what it got, plus a footer when there is more. `--mark` — everything read, unseen — is the one spelling that still uses `mailroomInbox`, now documented as exactly that. `--json` gains `remaining`, present only when the answer is a page.
- The accepted-race comment is deleted from the Swift CLI and the shim; the shim sends the same single request and renders the same footer, byte-equal.

### From the #293 review (both land here, as the review asked)

- **The version-skew blocker.** A 0.1.63 CLI against a #293 daemon printed "no unread posts" off a snapshot that carries no posts and then sent `mailroomInbox`, which moved its cursor to the room's latest — 182 posts unreachable, permanently. `GraphCommand.mailroomInbox` now moves **no cursor at all**: the daemon answers it with an error naming the skew ("this graphcode CLI predates the daemon's mailbox — nothing was marked read. Upgrade graphcode…"), so a client that received nothing advances nothing and the failure is loud and recoverable. The new CLI never sends it; `--mark` walks the unread mail through the mailbox page by page, headlines only, and reports the last id it passed.
- **`highestDeliveredID` under `search`** (and every other shape). It named the last *matching* post while unread posts below it were filtered out. It is now the highest id below which **every** unread post was handed over — `nil` for an empty page, `nil` when the first unread post was filtered out, exactly the matched prefix otherwise, and a page's edge only when nothing below it was skipped. Belt and braces: the daemon **refuses** `advanceCursor` together with `search` (`MailboxRefusal`, to the asker alone), since the protocol is reachable by clients we do not ship. Tests cover every answer shape, the reviewer's case verbatim, the refusal, and that the legacy command leaves the cursor untouched.

## Verification

Gate: full Xcode suite (1622 tests in 167 suites, 0 failures, on the stack rebased onto main post-#291), swiftlint 0 errors, swift-format clean, `graphcoded` and `graphcode-cli` schemes build; SwiftPM `swift build` and `.build/debug/graphcode` pass locally.

Tests added: the page and its `remaining`; an older answer decoding with nothing remaining; the cursor stopping at the page and the next request bringing the rest, persisted once per advance and never broadcast (a `MSG_PEEK` on the connection finds nothing); refusals leaving the cursor untouched and never touching a plain read; the registry returning a refused advance to the asker; the CLI footer and `--json` shape; and the shim sending one request with `advanceCursor` and rendering a paged backlog byte-equal to Swift.

## Not in this PR

Coalescing presence-poll broadcasts (fix 5), #289's diagnostics (fix 6), and encoding once per broadcast (fix 3, after #291).

Refs #288

### DIFF
--- a/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
+++ b/GraphcodeKit/Sources/CLI/GraphcodeCommand.swift
@@ -104,14 +104,14 @@ public enum GraphcodeCommand: Equatable, Sendable {
       graphcode mail post <project-path> [--topic <t>] <notice…>
                            post a notice to the whole graph, for whoever comes next
       graphcode mail inbox <project-path> [--headlines] [--full] [--mark] [--json]
-                           read your unread mail and mark the room read. A large
-                           backlog prints as headlines on its own and says so —
-                           --full insists on every body, --headlines insists on
-                           triage lines (deep-read either with `read`). --mark
-                           advances the cursor without printing the backlog, --json is
-                           the machine-readable shape. Combined, the output wins in the
-                           order --json > --mark > --headlines > --full; the cursor
-                           advances whichever flags you pass
+                           read your unread mail; what prints is what is marked read.
+                           A large backlog prints as headlines on its own and says
+                           so — --full insists on every body, --headlines insists on
+                           triage lines (deep-read either with `read`) — and comes a
+                           page at a time: run it again for the next page. --mark
+                           marks everything read without printing it, --json is the
+                           machine-readable shape. Combined, the output wins in the
+                           order --json > --mark > --headlines > --full
       graphcode mail read <project-path> <post-id>
                            one post in full — the deep-read half of --headlines
       graphcode mail list <project-path> [--search <text>] [--json]
@@ -866,9 +866,14 @@ extension GraphcodeCommand {
       if let search, !search.isEmpty {
         return unread ? "no unread posts match '\(search)'" : "no posts match '\(search)'"
       }
-      return unread
-        ? "no unread posts"
-        : "the room is empty — post one: graphcode mail post <project-path> <notice…>"
+      guard unread else {
+        return "the room is empty — post one: graphcode mail post <project-path> <notice…>"
+      }
+      return mailbox.prunedUnread > 0
+        ? "no unread posts — but \(mailbox.prunedUnread) landed since your last inbox and "
+          + "were pruned before you read them; the room keeps \(Mailroom.maxNotices) notices "
+          + "and \(Mailroom.maxLetters) letters, read it more often"
+        : "no unread posts"
     }
     let triaged = mailbox.bodiesTrimmed && !headlines
     let label = unread ? "mailroom, unread" : "mailroom"
@@ -883,6 +888,23 @@ extension GraphcodeCommand {
       lines.append(
         headlines || mailbox.bodiesTrimmed ? "  \(renderHeadline(post))" : "  \(render(post))")
     }
+    if unread, mailbox.prunedUnread > 0 {
+      // Said before the posts, in words, or the loop believes it is caught up: mail
+      // that landed after its cursor and was pruned before it asked is gone for good.
+      lines.append(
+        "  \(mailbox.prunedUnread) post\(mailbox.prunedUnread == 1 ? "" : "s") landed since your "
+          + "last inbox and \(mailbox.prunedUnread == 1 ? "was" : "were") pruned before you read "
+          + "\(mailbox.prunedUnread == 1 ? "it" : "them") — the room keeps "
+          + "\(Mailroom.maxNotices) notices and \(Mailroom.maxLetters) letters; read it more often")
+    }
+    if mailbox.remaining > 0 {
+      // A page, not the whole backlog: the cursor stopped at the last post above, so
+      // the same command again is the next page — said in words, or a loop would
+      // take one page for the lot.
+      lines.append(
+        "  \(mailbox.remaining) more unread past #\(mailbox.highestDeliveredID ?? 0) — "
+          + "run the same command again for the next page")
+    }
     return lines.joined(separator: "\n")
   }
 
@@ -894,8 +916,15 @@ extension GraphcodeCommand {
     struct Board: Encodable {
       var posts: [MailroomPost]
       var lastRead: Int?
+      /// Present only when the answer is a page, so the shape scripts already parse
+      /// is untouched until there is something to say.
+      var remaining: Int?
+      var prunedUnread: Int?
     }
-    let board = Board(posts: mailbox.posts, lastRead: mailbox.lastRead)
+    let board = Board(
+      posts: mailbox.posts, lastRead: mailbox.lastRead,
+      remaining: mailbox.remaining > 0 ? mailbox.remaining : nil,
+      prunedUnread: mailbox.prunedUnread > 0 ? mailbox.prunedUnread : nil)
     let encoder = JSONEncoder()
     encoder.outputFormatting = [.sortedKeys]
     encoder.dateEncodingStrategy = .iso8601
--- a/GraphcodeKit/Sources/GraphStore.swift
+++ b/GraphcodeKit/Sources/GraphStore.swift
@@ -517,7 +517,10 @@ public actor GraphStore {
 
   // MARK: - Commands
 
-  public func handle(_ command: GraphCommand) async {
+  /// `from` is the connection the command arrived on, when the registry knows it — what
+  /// lets a refusal meant for one client go to that client alone. Tests drive the
+  /// store without one and hear refusals through `onAnnounceError`.
+  public func handle(_ command: GraphCommand, from connectionID: UUID? = nil) async {
     // A loop inside a composite addresses itself by its own id — its briefing tells it
     // to `node memo <project> <its-own-id>`, and ids are unique across the whole tree,
     // so a caller has no reason to know how deep its target sits (the same rule
@@ -648,8 +651,8 @@ public actor GraphStore {
     case .mailroomPost(let text, let topic, let from):
       await mailroomPost(text: text, topic: topic, from: from)
 
-    case .mailroomInbox(let from):
-      mailroomInbox(from: from)
+    case .mailroomInbox:
+      refuseLegacyInbox(to: connectionID)
 
     case .mailroomWatch(let on, let topic, let from):
       mailroomWatch(on: on, topic: topic, from: from)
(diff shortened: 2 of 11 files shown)


scgopi:
## Independent review — final verdict at `7d8695c`

Reviewed in my own worktree (`review/295-audit`, branched from the PR head, private `-derivedDataPath`), as the delta on top of #293. I followed the branch across four force-pushes and re-ran at each: `6fa9f30` → `1610866` → `ddddbd6` → `72540c4` → `7d8695c`.

**Verdict: merge — after you put back the two tests `7d8695c` deleted.** Everything else I found is non-blocking. The three fixes are real; I attacked each of them with a probe rather than reading the diff, and they hold.

---

### ✅ What I verified, by probe, at `7d8695c`

`graphcode/Tests/MailboxCursorReviewTests.swift` on `review/295-audit`. Four of five pass; the fifth is finding 2 below.

| Claim | How I attacked it | Result |
|---|---|---|
| `search` + `advanceCursor` cannot eat unread mail | Encoded the pair through `JSONEncoder`/`JSONDecoder` — it survives the wire intact — then handed the **decoded** query to `GraphStore.mailbox` | ✅ `MailboxRefusal`, cursor still `nil`, all ten posts still unread. The guard is genuinely daemon-side, ahead of the room-off and identity guards; no caller convention is load-bearing |
| `highestDeliveredID` is honest in every answer shape | Empty answer; caught up; cursor past the end; a search whose first hit is not the first unread post; the whole room in one answer; **a room pruned under a reader whose cursor sits below what survived** | ✅ `nil` wherever nothing was handed over; after a prune it names the surviving high-water mark and counts the gap in `prunedUnread` |
| The legacy refusal reaches the asker, not everyone | Two connections registered on one store, one sends `.mailroomInbox`, the bystander reads with a **blocking** 1 s deadline (a `MSG_DONTWAIT` read here races `OutboundChannels` and gives a false pass — my own first probe did exactly that) | ✅ bystander gets nothing. This was a real defect at `72540c4` and `refuseLegacyInbox(to:)` fixes it correctly |
| A page can never strand a backlog | Served a room at its retention cap to a reader at cursor 0 | ✅ `remaining == 0`, whole surviving room in one answer, cursor at its last id |
| The cursor write is persisted, not broadcast, and clients still converge | Your `MSG_PEEK` test, plus reading the wiring: `ProjectRegistry`'s `onGraphChanged` is `saveGraph` + `refreshAwakeAssertion` only; broadcast is `notifyClients`, untouched | ✅ and convergence holds. `renderStatusMailroomLine` is the only reader of `lastMailroomRead`, and every `graphcode status` is a fresh connection served from the daemon's authoritative graph. The app rail uses per-window `seenMailroomPostID`, not the node cursor. Nothing on the wire lets a client write a whole graph back, so a stale in-app copy cannot clobber the cursor |
| Shim parity | Read `cliShimSource` against the Swift renderers line by line; extracted the shim as delivered and parsed it | ✅ both new footers and both `--json` keys are byte-identical, and `--mark`'s walk mirrors the Swift `repeat`/`while` |

Gate in my worktree at `7d8695c`: **113 of 114 pass** across `MailboxTests`, `MailroomTests`, `MailroomCommandTests`, `MailroomFetchTests`, `MailroomBudgetTests`, `MailroomLegacyNamesTests`, `RemoteCLIShimTests`, `BoardsOffRegressionTests` and my own suite — the one failure is my probe for finding 2. `-scheme graphcode-cli` and `-scheme graphcoded` both `** BUILD SUCCEEDED **`. `swiftlint lint`: 0 errors. `swift format lint --recursive --strict` over `GraphcodeKit graphcode graphcode-cli graphcoded MailroomKit`: clean.

---

### 🔴 Finding 1 — `7d8695c` deletes the only test of the fix this round exists for

```
$ diff <(git show 72540c4:…/MailboxTests.swift | grep -oE 'func [a-zA-Z]+\(' | sort) \
       <(git show 7d8695c:…/MailboxTests.swift | grep -oE 'func [a-zA-Z]+\(' | sort)
< func aSearchedInboxIsRefusedTheAdvance(
< func theLegacyInboxCommandMovesNoCursor(
```

Two removed, none added. After that:

- **`search` + `advanceCursor` has no test anywhere in the repo.** `grep -rn "MailboxRefusal" graphcode/Tests` finds only `theAdvanceIsRefusedWhereTheOldCursorCommandWas`, which covers room-off and unknown-reader — not the search pair. The one defect this round was opened to fix now ships with zero regression cover.
- **The legacy refusal has no test at all.** `grep -rn "predates the daemon" --include='*.swift' .` matches `GraphStore.swift:1758` and nothing else. Neither the message nor the per-connection routing of `refuseLegacyInbox(to:)` is asserted anywhere. (I was told a `MSG_PEEK` bystander test had landed with it; it has not — that is why I checked rather than relaying it.) The only surviving half is `MailroomTests.syncAdvancesCursorAndNeverMovesItBackward`, which asserts the cursor does not move.

Both deletions look collateral: `advancingTheCursorStopsAtWhatWasHandedOver` had to be rewritten (correctly) because `inboxPageSize = maxNotices + maxLetters` made its 405-post fixture unconstructible, and the two neighbours went with it. Nothing about the *behaviour* regressed — my probes prove that. But this PR's own history is a landmine that only a reviewer caught, and the fix for it is now untested. That is how it comes back.

Fix: restore the two tests, or cherry-pick `MailboxCursorReviewTests.aSearchedInboxIsRefusedAndMovesNothing` and `theLegacyRefusalMustNotBeBroadcastToBystanders` from `review/295-audit` — the second is worth having on its own, since nothing else pins "to the asker, not the room". Note the blocking-read helper in it: a `MSG_DONTWAIT` read straight after `handle` passes whether or not the bug is there.

---

### 🟡 Finding 2 — `prunedUnread` is silent for a reader that has never read

```swift
if let lastRead {
  let latest = posts.map(\.id).max() ?? lastRead
  pruned = max(0, latest - lastRead - selected.count)
}
```

A loop created by `graphcode node create` has `lastMailroomRead == nil` and keeps it until its first `mail inbox`. The branch never runs, so mail that landed **during its life** and was pruned before it read is reported as nothing lost — the exact silence `prunedUnread` exists to break.

Failing probe (`aReaderThatNeverReadIsStillToldWhatItLost`): loop created, 260 posts land, retention prunes 60, the loop runs its first inbox.

```
✘ Expectation failed: (answer.prunedUnread → 0) == 60
   "60 posts landed while this loop existed and were pruned before it read; it was told 0"
```

Ordinary path: a loop created, working for a few hours, reading the room late. The docstring says *"Zero for a reader the room does not know"* — but this reader **is** one the room knows; it has simply not read yet, and `nil` cannot tell those two apart.

The clean fix is upstream of `serve`: stamp `lastMailroomRead` with the room's latest id when a node is created. A live loop then never carries `nil`, "unread" always means "since you existed", and a new loop's first inbox stops being the entire retained backlog of a graph it was not part of. Not a merge blocker — an issue.

---

### On the paging path now being unreachable

Asked for explicitly, so plainly: `pruned()` caps the room at `maxNotices + maxLetters = 400` and `inboxPageSize` is now exactly that, so `selected.count > inboxPageSize` is structurally impossible. `remaining` is always `0`, the footer can never print, and `--mark`'s loop always exits after one pass.

**Keep it** — but do two things, because "unreachable" and "untested" are not the same problem and only one of them is forced:

1. **The paging path is not untestable, only untestable through the store.** `Mailroom.serve` takes `posts` as a parameter, so handing it 500 synthetic posts exercises the slice, `remaining`, `highestDeliveredID`'s page edge, the CLI footer and the shim footer without a room that can hold them. `aPagedBacklogRendersByteEqualToo` already does exactly this and is the reason the footer strings are still covered. Add the `serve`-level page test back the same way and nothing here is dead-in-the-sense-that-matters — dead render code that nobody exercises is how a wrong string reaches a user two years later.
2. **Pin the invariant as an invariant.** `#expect(Mailroom.inboxPageSize >= Mailroom.maxNotices + Mailroom.maxLetters)` with the reason in the comment, so lowering the page is a red test rather than a silently reopened pruning window. You have this; keep it next to the constant.

Removing the mechanism instead would make `page >= room` load-bearing with nothing left to catch a later change, which is the worse trade — the field costs one `Int` on the wire and the code costs four lines. Keeping it and saying so in the comment (structurally zero today; exists so that lowering the page is survivable rather than silent) is the right call.

While you are there: **`--mark`'s walk has no iteration bound** (`repeat { fetch } while remaining > 0`, and `while True` in the shim). Its condition is now structurally always-false, which makes an unbounded loop against a socket whose peer version you do not control pure downside. `for _ in 0..<8` costs nothing.

---

### Nits

- **`Mailbox.highestDeliveredID`'s docstring is only true for `.unread`.** It promises *"the highest id below which **every** unread post was handed over"*, but `.board` and `.post` still return `selected.last?.id` — and `theCursorCeilingNeverPassesMailThatWasNotHandedOver` pins `serve(graph, .board, search: "release").highestDeliveredID == 3`, the highest *matching* post. Harmless today because `GraphStore` ignores `advanceCursor` for those selections, but the field's own doc is what the next caller reads, and the next caller may be the shim. One clause fixes it.
- **`refuseLegacyInbox` drops the refusal silently if the asker is not in this store's `connections`.** `if let connectionID, connections[connectionID] != nil` — a `.graphCommand` for a project this connection never opened still routes to the store (the registry guards on `stores[canonicalPath]`, not on membership), and the old CLI's `waitForEvent { .graphChanged, .errorOccurred }` would then hang rather than fail. Not reachable through the real client, which opens first — but the registry already has the `fileDescriptor` and could answer without the store, which removes the case entirely.
- **`--mark` ships the whole backlog only to discard it**: up to 400 headlined posts across the wire for a command that prints one line, in a PR about wire size. A `posts: false` on the query, or a `.markRead` selection, would make it one small round trip.
- **The shim hard-codes `MAX_NOTICES = 200` / `MAX_LETTERS = 200`** where Swift reads `Mailroom.maxNotices`/`maxLetters`. `aPagedBacklogRendersByteEqualToo` would catch drift, so this is a note — but a delivered shim is the one copy nobody rebuilds by hand.
- **Behaviour change worth a line in the PR body**: with the room off, `mail inbox` used to print the posts and *then* fail; the refusal now precedes the answer, so it fails with nothing printed. Reading was previously ungated on purpose ("a room switched off still shows what was said while it was on") and for the inbox it now effectively is not.

---

### For the record, on earlier SHAs

Two things I found that are already fixed, noted only because they bear on how much the gate can be trusted:

- **`ddddbd6` did not compile.** `MailroomCommandTests.swift:418: value of optional type 'Bool?' not unwrapped` — I hit it on my own first run there. No gate can have been green at that SHA. `72540c4` fixed it, and the fix flipped the assertion from `prunedUnread == nil` to `== 3`, because the old spelling was `#expect(try? decoder.decode(…).prunedUnread == nil)` — an `#expect` over an `Optional<Bool>`, which cannot fail. Worth grepping for that shape elsewhere in the suite.
- At `6fa9f30` all three of my opening probes failed: `search` + `advanceCursor` moved the cursor to `#10` and lost nine posts; `mailroomInbox` moved a cursor to `#140` having delivered nothing; and `inboxPageSize = 100` against a 200 + 200 retention cap let 100 posts be pruned between page one and page two. All three are properly closed now, and the third was closed by tying the page to retention rather than by making the page bigger, which is the right shape of fix.

---

### Summary

The cursor rule this arrives at is correct, and the fixes for the previous review's landmine survived everything I could throw at them — the refusal is in the daemon, not in a caller's manners, and `highestDeliveredID` is honest in every shape I could build, including one the PR does not test. Ship it. But put back the two tests that went missing on the way here: the fix for a bug a reviewer had to find should not merge with nothing watching it.

PR #308, opened 2026-09-06
Persist the room beside the graph, and write off the actor (#307)

### LINKED ISSUE #307: Every graph change synchronously writes 271 KB of mail bodies to disk, holding the GraphStore actor
Noticed as an aside by the `MailConcurrent` end-to-end loop (`node memo` taking 0.16–2.13 s, 20–100× an inbox read, present with and without a wedged client so **not** #288), then measured and root-caused here on the shipped **0.1.64-beta2**.

## The shape is #288 again, one layer over

`GraphStore.broadcast()`:

```swift
private func broadcast() {
  onGraphChanged?(graph)   // ← synchronous
  notifyClients()
}
```

`ProjectRegistry` wires that to `persistence.saveGraph(updatedGraph)` — a synchronous disk write of the entire graph — and `broadcast()` runs **on the `GraphStore` actor**. So every graph mutation holds the actor across a full serialize-and-write to disk, exactly as it used to hold the actor across a socket write before #291.

## The payload is what #293 was supposed to remove

#293 took the Mailroom out of the *wire* snapshot. It is still in the *persisted* one:

| | bytes |
|---|---|
| persisted project JSON | **323,603** |
| of which `mailroom` | **271,256** (330 posts) |
| **share** | **84%** |
| wire `graphChanged` frame, for contrast | 56,604 (mailroom = 68-byte digest) |

So the broadcast got 77% smaller and the disk write did not move at all. Every memo, every post, every persisted state change re-serializes 271 KB of mail bodies that did not change.

## Measured cost

`graphcode node memo`, three consecutive calls on an otherwise quiet daemon:

```
real 0.03
real 0.77
real 0.87
```

`MailConcurrent` saw up to **2.13 s** under load. A socket-level `openProject` round trip on the same daemon is **0.003 s**, so a memo costs 10–700× a read, and the variance points at the filesystem rather than at compute.

Every loop writes memos — it is how a pass leaves anything behind for the next one — so this is on a hot path for the whole graph, and it gets slower as the room fills.

## Two independent fixes, either helps

1. **Do not hold the actor across the write.** `onGraphChanged` should hand the snapshot to a serial writer and return, the way `OutboundChannel` now takes frames. Coalescing consecutive saves would also collapse a burst of memos into one write.
2. **Stop persisting the room inside the project graph**, or persist it separately so a node change does not rewrite every post. #293 established that posts do not belong in the snapshot clients receive; the same argument applies to the file rewritten on every mutation.

The first is the direct analogue of #291 and fixes the stall regardless of payload size. The second is the analogue of #293 and removes 84% of the bytes.

## Why it was not caught

Every measurement in the #288 series was taken at the socket. The disk half was never instrumented, and PR #303's diagnostics would time `persistence` as a phase — which is precisely how this would have surfaced on its own.

Related: #288, #291, #293, #289

scgopi:
## The defect

`GraphStore.broadcast()` called `persistence.saveGraph(graph)` synchronously, on the actor, and `saveGraph` serialised the whole graph — **323,603 bytes on the graph that filed #307, 271,256 of them (84%) mail bodies that had not changed**. So every memo, state tick and cursor move held the `GraphStore` actor across a full encode-and-write: a `node memo` measured at 0.03–2.13 s against a 0.003 s socket round trip, with the variance coming from the filesystem. It is #288's shape one layer over, and the payload #293 took out of the wire snapshot was still in the file.

## The change — both halves the issue names, each sufficient on its own

**The room lives beside the graph, not in it.** `ProjectPersistence.saveGraph` writes the graph *without* its posts to `<name>.json` (rewritten on every change) and the room to `<name>.mailroom.json`, rewritten only when the room's `MailroomDigest` changed since the last write — so a memo rewrites ~52 KB, not 323 KB, and a post rewrites the room once. `loadGraph` reads both, the sidecar winning; a graph file saved before the split still carries its posts inline and loads exactly as it did. `deleteGraph` removes both. The same argument #293 made for the wire, applied to the file.

**The write leaves the actor.** `ProjectRegistry` hands each snapshot to a `GraphWriter` — one serial queue, latest snapshot per project wins — and the store's `onGraphChanged` returns at once; a burst of memos collapses to one write. `GraphWriter.flush()` waits for everything queued: the daemon calls it on both exit paths (SIGTERM/SIGINT source, the binary-replaced timer) so a change applied a moment before a restart is on disk, and `ProjectRegistry(persistsSynchronously: true)` flushes on every save for the tests that read the file straight after a command.

Downgrade note: a daemon older than this reads `<name>.json` only, so on a downgrade the room would appear empty until the next upgrade — the posts are untouched in the sidecar. Same class as #298, one direction only, and named here so it is not a surprise.

## Verification

Gate: full Xcode suite (gate on this head in flight; numbers and exit codes in a comment when it completes), swiftlint 0 errors, swift-format clean, `graphcoded` and `graphcode-cli` schemes build; SwiftPM `swift build` and `.build/debug/graphcode` pass.

`MailroomPersistenceTests`: the graph file carries neither a `mailroom` key nor a post body (a nonce planted in a post never appears in it) while the sidecar does; a change that leaves the room alone leaves the sidecar's mtime alone; loading merges both; deletion removes both; a pre-split graph file with inline posts loads them; a burst of fifty saves lands the newest once after `flush`. `ProjectRegistryTests` and `DuplicateProjectPathTests` run with synchronous persistence and are otherwise unchanged.

Closes #307.

### DIFF
--- a/GraphcodeKit/Sources/GraphWriter.swift
+++ b/GraphcodeKit/Sources/GraphWriter.swift
@@ -35,6 +35,20 @@ public final class GraphWriter: @unchecked Sendable {
     queue.async { [self] in drain() }
   }
 
+  /// The newest snapshot of a project — the one still queued, if there is one, else
+  /// the file. Every reader of the persisted graph goes through here rather than
+  /// through the file: a save that has left the actor and not yet reached the disk is
+  /// otherwise invisible, and a delete of a *closed* project (no live store) that read
+  /// the file to find the loops whose sessions it must end would end fewer than exist
+  /// and leave the rest running.
+  public func load(path: String) -> LoopGraph? {
+    lock.lock()
+    let queued = pending[path]
+    lock.unlock()
+    if let queued { return queued }
+    return persistence.loadGraph(path: path)
+  }
+
   /// Returns once everything queued so far is on disk.
   public func flush() {
     queue.sync { drain() }
--- a/GraphcodeKit/Sources/GraphWriter.swift
+++ b/GraphcodeKit/Sources/GraphWriter.swift
@@ -1,0 +1,56 @@
+import Foundation
+
+/// Persists graphs off the actor that changes them — the disk-side twin of
+/// `OutboundChannel` (issue #307).
+///
+/// `GraphStore.broadcast()` used to call `persistence.saveGraph` synchronously, so every
+/// mutation held the `GraphStore` actor across a full serialise-and-write — the shape
+/// #291 removed from the socket path, one layer over: a memo measured at 0.03–2.13 s
+/// against a 0.003 s socket round trip, with the variance coming from the filesystem.
+///
+/// A save is handed here and the actor returns. One serial queue writes; consecutive
+/// saves of the same project collapse to the newest snapshot (the graph is a value and
+/// the file is a whole, so nothing older has anything left to say), which turns a burst
+/// of memos into one write. `flush` waits for everything queued — what the daemon calls
+/// on its way out, and what a test calls before reading the file back.
+public final class GraphWriter: @unchecked Sendable {
+  private let persistence: ProjectPersistence
+  private let queue = DispatchQueue(label: "dev.graphcode.graphcoded.persist", qos: .utility)
+  private let lock = NSLock()
+  private var pending: [String: LoopGraph] = [:]
+  private var scheduled = false
+
+  public init(persistence: ProjectPersistence) {
+    self.persistence = persistence
+  }
+
+  /// Queues the newest snapshot of a project and returns at once.
+  public func save(_ graph: LoopGraph) {
+    lock.lock()
+    pending[graph.project.path] = graph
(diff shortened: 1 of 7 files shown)


scgopi:
Gate on e51b80e, private DerivedData: `xcodebuild test` exit 0 — 1646 tests / 171 suites / 0 failures; `graphcode-cli` build exit 0; `graphcoded` build exit 0; swiftlint 0 errors; swift-format clean; Linux green on this head.

The merged-result failures were real and are fixed by read-through: `GraphWriter.load(path:)` returns the snapshot still queued for a project before falling back to the file, and every registry read — the delete of a *closed* project included, which reads the graph to find the sessions it must end — goes through it. The two tests stood in for a daemon restart inside one process; they now flush the first daemon's writer the way the daemon does on its way out, and the one that reads the file after every command asks for synchronous persistence. New test: a load returns the queued three-loop graph while the file still holds one.

Issue #306, opened 2026-09-06
A Claude Code loop idle over 60s never receives a watch wake or a --follow-up message


scgopi:
Found by the `MailWatcher` end-to-end loop against the shipped **0.1.64-beta2 (build 252)**, with millisecond-stamped `zmx` label logs and two independent loops showing the same flip.

## Symptom

**0 of 18** matching Mailroom posts were typed into a watching session. All 18 landed in its log as `follow-up staged` and none arrived during 6+ minutes idle.

The watch itself is fine: `mail watch --topic e2e-poster` was armed at 21:06:41Z; the 18 matching posts (`#338-#351`, `#354-#357`) all matched, and `#337` under the adjacent topic `e2e-poster-probe` was correctly **not** matched — so topic filtering is exact. A plain `node send` (no `--follow-up`) from another loop **did** reach the session live at 21:13:52Z. Only the follow-up path is dead.

## Cause

Three things compose into it:

1. Watch wakes ride `deliverAdHocMessage(followUp: true)`. `deliversLater()` stages when `presence != idle`; `drainPendingFollowUps()` delivers **only** on `presence == idle`.
2. Claude Code fires an `idle_prompt` **Notification** roughly 60 seconds after a turn ends.
3. graphcode's `Notification` hook (`PresenceHooks`) maps **every** notification to `presence = awaitingInput`, with no `idle_prompt` filter.

So a loop settles to `idle`, and 60 seconds later its own idle notification flips it to `awaitingInput` — where it stays. Staged follow-ups then wait for an `idle` that will not come again until the loop happens to run a turn and stop.

Measured, two loops, same shape:

```
MailWatcher: presence=idle 21:07:25  ->  awaitingInput 21:08:25  (60.1s)
MailPoster:  presence=idle 21:13:28  ->  awaitingInput 21:14:28  (60.0s)
```

## Why this matters more than it first looks

- **The Mailroom watch is effectively non-functional for Claude Code loops**, which is the default backend. The feature appears armed, reports itself armed, and silently delivers nothing.
- `node send --follow-up` has the same fate — the polite, non-interrupting way for loops to talk to each other is the one that does not arrive.
- It fails **silently**. The poster sees a successful post, the watcher sees an armed watch, and nothing says the delivery is parked. That is the same shape as every other defect in the #288 series.
- It plausibly explains loops that appear to "go idle and stop responding": they are in `awaitingInput`, not idle, and anything staged for them is waiting on a state they have left.

[...]

PR #309, opened 2026-09-07
Drain staged follow-ups once, in order, and let watch --off take its wakes (#304)

### LINKED ISSUE #304: mail watch --off leaves staged deliveries queued, and they can duplicate
Found by accident while cleaning up after a test loop, on the shipped **0.1.64-beta2 (build 252)**.

## What happens

A loop watching the Mailroom while it is busy has matching posts **staged to its memory** rather than typed in live — that is the documented design, the same path a `--follow-up` message takes. But:

1. `graphcode mail watch --off` removes the subscription and **does not cancel or drain what is already staged**.
2. Staged deliveries keep arriving for many minutes afterwards, in strict post order.
3. They arrive **even though the reader's cursor is already past them** — `lastMailroomRead = 353` while posts `#276`, `#277`, `#280`, `#281` were still being delivered.
4. At least one was delivered **twice** (`#280` arrived two times).

## Evidence

The watching node's persisted state after `--off` shows no watch at all:

```
PerfTriage mailroom-related keys: ['lastMailroomRead']
  lastMailroomRead = 353
```

So the subscription is genuinely gone from the graph, and the deliveries are coming from a queue that `--off` does not reach.

Sequence observed: `filler 1 of 70`, `2 of 70`, `3`, `4`, `5`, `5` (duplicate), `6` … continuing after the watch was turned off twice, re-scoped to a different topic in between, and the inbox marked read to `#353`.

## Why it matters

- **There is no way to stop it.** `--off` is the only control, and it does not apply to in-flight deliveries. A loop that accidentally subscribes to a busy topic will be fed that backlog into its context regardless of what its operator does — and for an agent session, that is context budget spent on posts it has already read.
- **The duplicate is the more serious half.** Delivery is supposed to be at-most-once into a session; a post arriving twice means a watcher cannot treat delivery as a signal to act exactly once.
- Redundant with the cursor: delivering posts below `lastMailroomRead` contradicts what the cursor is for.

## Suggested behaviour

- `--off` should drop staged-but-undelivered posts for that watcher, not just the subscription.
- A staged post whose id is at or below the reader's `lastMailroomRead` should be dropped rather than delivered.
- Delivery should be idempotent per post id per watcher.

## Related

This is the *delivery* half of the Mailroom, which #293 did not touch — that PR rewrote how mail is **read**. Every review in the #288 series concentrated on the read path, so delivery has had no scrutiny. `MailWatcher` was created to test exactly this and found the ordering side; this is the teardown side.

Related: #288, #293

scgopi:
## The defect

`drainPendingFollowUps` — the delivery of staged `--follow-up` messages and Mailroom watcher wakes to a loop once it goes idle — iterated a copy of the pending list, awaited a presence read and a delivery *per item*, and finished by **assigning** `pendingFollowUps = remaining`. The presence poll calls it every fifteen seconds, so a second drain routinely started while the first was suspended mid-delivery. Two overlapping drains both delivered the same items (the duplicates #304 saw), and the later one's final assignment overwrote the earlier one's — dropping whatever had been appended meanwhile and whatever the other drain had retained (the 5 of 18 that never arrived, and the reordering). Separately, `mail watch --off` removed the subscription but not the wakes already staged, which then arrived for minutes — past the reader's cursor.

## The change

- **Non-reentrant, take-and-clear.** A drain in flight makes the next call return; the drain takes the whole pending list and clears it in one actor step, delivers from the local batch, and puts its retries back *in front of* whatever was queued while it ran. Nothing is ever overwritten; order is kept.
- **Watcher wakes are tagged with their post** (`PendingFollowUp.watchedPostID`). `mail watch --off` drops the wakes still staged for that watcher — a peer's `--follow-up` message to the same loop is not a wake and stays. And at delivery time a wake is only owed while the watch stands and the post is still unread: a reader whose cursor has passed the post (an inbox in between) is not woken for it.

## Verification

Gate: full Xcode suite (gate on this head in flight; numbers and exit codes in a comment when it completes), swiftlint 0 errors, swift-format clean, `graphcoded` and `graphcode-cli` schemes build; SwiftPM build and CLI smoke pass.

`FollowUpDrainTests`: two polls overlapping on a slow idle reading deliver three staged messages exactly once each and in order, and a fourth staged during the overlap is delivered once afterwards (fails on `main`: duplicates and a loss); `--off` drops the staged wake and still delivers the peer's follow-up; a wake for a post the reader has since read is not sent.

Related: #306 (why the queue fills), #288.

Closes #304.

### DIFF
--- a/GraphcodeKit/Sources/GraphStore.swift
+++ b/GraphcodeKit/Sources/GraphStore.swift
@@ -173,7 +173,24 @@ public actor GraphStore {
   /// presence poll. The content is in the target's memory log from the moment it was
   /// queued, so losing this queue to a restart delays the message to the next wake
   /// rather than dropping it.
-  private var pendingFollowUps: [(nodeID: UUID, text: String)] = []
+  /// A deferred delivery: a `--follow-up` message, or a Mailroom watcher's wake for one
+  /// post (`watchedPostID`), waiting for its target to go idle.
+  struct PendingFollowUp: Equatable {
+    let nodeID: UUID
+    let text: String
+    /// The post a watcher's wake is about — what lets `mail watch --off` drop the wakes
+    /// still queued, and lets a wake for a post the reader has since read go unsent.
+    /// `nil` for a message a peer or a human sent.
+    let watchedPostID: Int?
+  }
+
+  private var pendingFollowUps: [PendingFollowUp] = []
+  /// `drainPendingFollowUps` runs across several awaits, and the presence poll that
+  /// calls it runs every fifteen seconds: without this, a second drain started while
+  /// the first was suspended delivered the same items again and, on finishing, wrote
+  /// its own idea of what remained over the first's — dropping whatever had been
+  /// queued in between (issue #304: duplicates, lost mail, and out-of-order delivery).
+  private var isDrainingFollowUps = false
 
   /// A poller holds `self` weakly, so a store going away already stops it *doing*
   /// anything — but the task itself keeps sleeping in its loop forever. Harmless for
@@ -1682,7 +1699,8 @@ public actor GraphStore {
         "mailroom — new post #\(post.id)\(topicSuffix(post)) from \(post.author): "
         + "\(preview) — read it with: graphcode mail inbox \(graph.project.path)"
       await deliverAdHocMessage(
-        to: node.id, text: nudge, from: nil, followUp: true, mirror: false)
+        to: node.id, text: nudge, from: nil, followUp: true, mirror: false,
+        watchedPostID: post.id)
     }
   }
 
@@ -1824,6 +1842,9 @@ public actor GraphStore {
         recordMemory(watcherID, "mailroom: stopped watching")
       }
       graph.nodes[id: watcherID]?.mailroomWatch = nil
+      // What `--off` is for: the wakes already staged for this watcher go with the
+      // watch. A peer's `--follow-up` message to the same loop is not a wake and stays.
+      pendingFollowUps.removeAll { $0.nodeID == watcherID && $0.watchedPostID != nil }
     }
   }
 
@@ -2601,7 +2622,7 @@ public actor GraphStore {
   /// it landed is the one wrong answer.
   private func deliverAdHocMessage(
     to nodeID: UUID, text: String, from senderID: UUID?, followUp: Bool = false,
-    mirror: Bool = true
+    mirror: Bool = true, watchedPostID: Int? = nil
   ) async {
     guard let target = graph.nodes[id: nodeID] else {
       announceError("message not delivered: no loop \(nodeID) in this graph")
@@ -2633,7 +2654,8 @@ public actor GraphStore {
     // dropping it.
     if followUp, deliversLater(to: target) {
       recordMemory(nodeID, "follow-up staged: \(message)")
-      pendingFollowUps.append((nodeID: nodeID, text: message))
+      pendingFollowUps.append(
+        PendingFollowUp(nodeID: nodeID, text: message, watchedPostID: watchedPostID))
       return
     }
 
@@ -2698,10 +2720,23 @@ public actor GraphStore {
   /// the reading this waits for. A target that resolved or died is simply dropped from
   /// the queue: its memory log has carried the message since it was staged.
   private func drainPendingFollowUps() async {
-    guard !pendingFollowUps.isEmpty else { return }
-    var remaining: [(nodeID: UUID, text: String)] = []
-    for pending in pendingFollowUps {
+    guard !pendingFollowUps.isEmpty, !isDrainingFollowUps else { return }
+    isDrainingFollowUps = true
+    defer { isDrainingFollowUps = false }
+    // Taken and cleared in one actor step, delivered from the local batch. Anything
+    // queued while a delivery below is suspended lands in `pendingFollowUps` untouched
+    // and is folded back in behind the retries at the end — never overwritten.
+    let batch = pendingFollowUps
+    pendingFollowUps = []
+    var remaining: [PendingFollowUp] = []
+    for pending in batch {
       guard let node = graph.nodes[id: pending.nodeID], !node.isResolved else { continue }
+      // A watcher's wake is only owed while the watch stands and the post is unread:
+      // `mail watch --off` after the wake was staged, or an inbox that has since read
+      // past the post, means the wake has nothing left to say.
+      if let postID = pending.watchedPostID {
+        guard node.mailroomWatch != nil, (node.lastMailroomRead ?? 0) < postID else { continue }
+      }
       switch MessageBus.deliverability(to: node) {
       case .targetBusyWithACheck:
         remaining.append(pending)
@@ -2723,7 +2758,7 @@ public actor GraphStore {
       }
       _ = await deliverToSession(node, pending.text)
     }
-    pendingFollowUps = remaining
+    pendingFollowUps = remaining + pendingFollowUps
   }
 
   private func announceError(_ message: String) {
--- a/graphcode/Tests/FollowUpDrainTests.swift
+++ b/graphcode/Tests/FollowUpDrainTests.swift
@@ -1,0 +1,156 @@
+import ComposableArchitecture
+import Foundation
+import GraphcodeKit
+import MailroomKit
+import Testing
+
+#if canImport(Darwin)
+  import Darwin
+#endif
+
+/// Issue #304: the drain of staged follow-ups runs across awaits and the presence poll
+/// re-enters it every fifteen seconds. Two drains overlapping delivered the same items
+/// twice and the later one's bookkeeping overwrote the earlier's — mail duplicated, mail
+/// lost, mail out of order. And `mail watch --off` left the wakes already staged to
+/// arrive anyway, past the reader's cursor.
+@Suite
+struct FollowUpDrainTests {
+  /// What the target's session appears to be doing; the drain asks each time.
+  private actor Readings {
+    var presence: Presence = .busy
+    var delay: Duration = .zero
+    func set(_ presence: Presence, delay: Duration = .zero) {
+      self.presence = presence
+      self.delay = delay
+    }
+    func read() async -> PresenceReading {
+      if delay > .zero { try? await Task.sleep(for: delay) }
+      return PresenceReading(presence: presence, confidence: .reported)
+    }
+  }
+
+  /// A connection with a channel lifecycle of its own and a reader draining it: a bare
+  /// `/dev/null` descriptor can be dropped as "disconnected" when its number is reused
+  /// across the suite, and then the poll never drains.
+  private final class Attachment: @unchecked Sendable {
+    let daemonEnd: Int32
+    private let peer: Int32
+    private let drainer: Task<Void, Never>
+    init() {
+      var pair: [Int32] = [0, 0]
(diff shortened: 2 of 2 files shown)


scgopi:
Gate on 7478915, private DerivedData: `xcodebuild test` exit 0 — 1651 tests / 173 suites / 0 failures; `graphcode-cli` build exit 0; `graphcoded` build exit 0; swiftlint 0 errors; swift-format clean; Linux green on this head (with the CLI smoke).

Fail-without / pass-with, on the branch base's `GraphStore`: without the fix, two overlapping polls delivered each of three staged messages **three** times, out of order (`staged 1, staged 2, staged 1`), and a wake for an already-read post went out — the duplicate, the loss and the reordering #304 measured, from the one `pendingFollowUps = remaining` overwrite. With it: once each, in order, a message staged mid-overlap once afterwards, `--off` drops its wakes and keeps a peer's follow-up, a read post is not woken for.

Not in scope: #306 (why the queue fills) — this makes the drain correct, not less starved.

scgopi:
Head moved to 79dd8d9 — test-only: the drain tests attached a bare `/dev/null` descriptor for the poll's connection, which the suite's descriptor reuse can drop as "disconnected" (it bit the sibling #310's e2e test in a full run). They now attach through a socketpair with a reader draining it and close the daemon end through `OutboundChannels`, the pattern every stable socket test here uses. Source unchanged. Re-gating; numbers to follow.

scgopi:
Gate on 79dd8d9, private DerivedData: `xcodebuild test` exit 0 — 1651 tests / 173 suites / 0 failures; `graphcode-cli` build exit 0; `graphcoded` build exit 0; swiftlint 0 errors; swift-format clean; Linux green on this head.

scgopi:
## Independent review — **I would merge this**, with one small follow-up and one caveat about #306

Reviewed in my own worktree (`worktrees/review-309`) on `7478915ae207147bd4435af59569ba2ec3579ca6`, confirmed equal to `refs/pull/309/head`. I did not trust the combined test — I wrote five of my own and ran them against **both** this head and `origin/main`'s `GraphStore.swift` in the same worktree, so every claim below is a measured before/after rather than a reading.

### Gate (private DerivedData, `xcodebuild`'s own exit code on the line after each)

| Step | Exit | Result |
|---|---|---|
| `xcodebuild -scheme graphcode … test` | `0` | `Test run with 1651 tests in 173 suites passed`, `** TEST SUCCEEDED **` |
| `xcodebuild -scheme graphcode-cli … build` | `0` | ✅ |
| `xcodebuild -scheme graphcoded … build` | `0` | ✅ |
| `swiftlint lint` | `0` | 0 `error:` |
| `swift format lint --recursive --strict` | `0` | clean |
| `scripts/cli-smoke.sh` | `0` | five verbs ok against a throwaway daemon |
| Linux build | pass | [run 34070361927](https://github.com/scgopi/GraphCode/actions/runs/34070361927/job/101586438199) |

> ⚠️ Trap for the next reviewer, not a defect here: `worktree-build-check` tells you to symlink `.build` to the main checkout, and `cli-smoke.sh` guards its `swift build` on `[ -x "$bin/graphcode" ]`. So the smoke run silently exercised a **Sep-1 binary from the shared `.build`** and reported three false `FAIL`s (`unknown command: mail`). Replacing the symlink with a real dir and running `swift build` gave the exit 0 above. The two skills collide.

### The three symptoms, verified separately — all three genuinely fixed

Your tests are not vacuous. Checking out `origin/main`'s `GraphStore.swift` under the same test files, **all three of yours fail**, and so do two of mine that pass on this head:

| Property | On `main` | On this head |
|---|---|---|
| **Exactly once** — 3 staged + 1 staged mid-overlap | **12 deliveries for 4 messages** (each ×3) | exactly 4 ✅ |
| **In order** under overlap | `[1, 2, 1, 1, 3, 2, 2, late, 3, 3, late, late]` | `[1, 2, 3, late]` ✅ |
| **Staged during an overlapping drain** | duplicated, order broken | arrives **once, last** ✅ |
| `--off` racing an **in-flight** drain (queue already taken into the local batch) | wake delivered **twice** | dropped ✅ |
| Wake for an already-read post | delivered | not sent ✅ |

That last row matters: `--off`'s `removeAll` cannot see a batch a drain has already taken, and I expected that to leak. It does not, because the delivery-time guard re-checks `mailroomWatch != nil` per item. Belt and braces, and the braces are the ones that hold.

**Starvation:** retries-in-front cannot starve a newly staged message. Each drain takes the *whole* list — retries and new alike — and visits every element; front-placement changes order, not reachability. ✅

**In-flight guard exit paths:** the `defer` is registered *after* the `guard` and after the flag is set, so an early return never clears a flag it did not set; nothing in the drain throws, and Swift cancellation does not unwind an `async` function. The flag clears on every path that returns. No wedge from a throw. ✅

### What I could still break — two ordering holes survive, so "in order" in the title is overstated

Both are **pre-existing** (they fail on `main` identically, so neither is a regression), but both produce #304's *stated* reordering symptom, so I don't think #304 should be closed as fully fixed without a word about them.

**1. A later follow-up jumps the whole queue.** Two different readings decide "deliver now": `deliversLater` consults the **cached** `node.presence` the poll last wrote, while the drain takes a **live** reading per item. When they disagree — poll saw idle, session is mid-tool-call when the drain asks — three messages already queued are retried while a fourth staged afterwards bypasses the queue and is typed in first. Deterministic, no overlap involved:

```
delivered == ["staged late"]   // while "staged 1", "staged 2", "staged 3" sit queued
```

**2. A turn ending mid-drain reorders the same loop's queue.** The presence read is *per item*, so busy for item 1 and idle for item 2 delivers 2 and 3 and retries 1 — and retries-in-front then delivers 1 on the *next* drain, after the two that overtook it:

```
delivered == ["staged 2", "staged 3", "staged 1"]
```

That is literally #304's "`#357` delivered before `#347`", reached with a single drain, so the in-flight guard cannot see it. Cheap fix: once an item for a node goes to `remaining`, `continue` past the rest of *that node's* items in this drain, and have `deliversLater` return true whenever anything is already queued for the target.

### One real gap in the `--off` half: re-scoping keeps the abandoned topic's wakes

`--off` drops staged wakes; `watch --on --topic other` does not — the `removeAll` lives only in the `else` branch, and the delivery-time guard asks whether *a* watch stands, never `watch.matches(post.topic)`. #304 records the watch being "re-scoped to a different topic in between" as part of the observed sequence, so this is squarely in scope:

```
// watch --topic alpha; post to alpha; watch --topic beta; go idle
delivered contains "an alpha post"   // ✗ — wake for a topic no longer watched
```

The guard already has the post id and `graph.mailroom` still holds the post, so this is a small addition to the existing `if let postID` block rather than new state. I'd take this one before merge if it's easy; it's the only finding that's inside the PR's own claim surface.

### Two smaller notes

- **No timeout on the presence read.** `onReadPresence` → `PTYProcessSession.waitCollectingOutput()` has no deadline, and `ssh`'s `ConnectTimeout=10` bounds the connect, not a hung remote command. An `await` that never returns now holds `isDrainingFollowUps` forever and freezes the queue permanently. I checked whether the guard makes this *worse* than `main` and it does not — `main` re-enters into the same hang and additionally re-delivers everything before the stuck item — but a bounded read is worth a follow-up, because "frozen" and "working" look identical from outside.
- **Latency.** A message staged during a drain now waits for the next drain (≤15 s poll, or the next command's `drainAndBroadcast`) rather than being picked up by the re-entrant one. Against #304's measured 511–835 s lag that is noise; a bounded `while !pendingFollowUps.isEmpty` re-loop at the end would remove it entirely.
- nit: the queue's doc paragraph ("Queued rather than delivered… the content is in the target's memory log…") now attaches to `struct PendingFollowUp` instead of to `pendingFollowUps`.

### Is this worth shipping while #306 is open?

**Ship it, but ship them together.** With #306 open this fix is correct and **unobservable on the default backend**: a Claude Code loop idle over 60 s sits at `presence == awaitingInput`, `MessageBus.deliverability` returns `.targetBusyWithACheck`, and the drain appends to `remaining` and `continue`s — it never reaches the presence read this PR governs. So on Claude Code loops nothing is delivered either way, and no field observation can distinguish this fix from `main`.

Merging it alone is still right: it is a strict improvement, it is what makes the backlog **safe to deliver at all** once #306 unblocks it, and landing the drain fix *after* #306 would mean shipping one beta in which a large stale queue drains through the buggy path — exactly the 18-post window that produced the duplicates and the 5 losses. Order matters: **#309 first or same beta, never #306 first.** But the release note shouldn't claim #304 is verified until a beta carries both and an end-to-end watcher measures 18 of 18, in order, once each.

<details>
<summary>The five probes I ran (drop into <code>graphcode/Tests/</code>, <code>tuist generate</code> first — new test files are invisible until you do)</summary>

Three fail on this head (the two ordering holes and the re-scope gap); all five fail on `main`. Shared harness: a scripted `onReadPresence` that is busy by default and hands out a queued list of readings for the target, so every case is deterministic with no `sleep`-based racing except the two that deliberately overlap two polls.

```swift
/// Two readings decide "deliver now": `deliversLater` reads the cached presence, the
/// drain reads live per item. When they disagree, a later message bypasses the queue.
@Test
func aLaterFollowUpDoesNotBypassOnesAlreadyQueuedForTheSameLoop() async {
  let (store, target, sender) = await makeStore(readings: readings, delivered: delivered)
  await stage(store, target, sender, ["staged 1", "staged 2", "staged 3"])
  await readings.scriptTargetReads([.idle])   // refresh caches idle; drain reads busy
  await store.pollPresence()
  #expect(delivered.value.isEmpty)
  await stage(store, target, sender, ["staged late"])
  #expect(texts(delivered) == [])             // actual: ["staged late"]
}

/// The turn ends between item 1 and item 2 of a single drain.
@Test
func aRetriedMessageIsNotOvertakenByALaterOneForTheSameLoop() async {
  let (store, target, sender) = await makeStore(readings: readings, delivered: delivered)
  await stage(store, target, sender, ["staged 1", "staged 2", "staged 3"])
  await readings.scriptTargetReads([.busy, .busy, .idle, .idle])
  await store.pollPresence()
  await readings.scriptTargetReads([.idle, .idle, .idle, .idle])
  await store.pollPresence()
  // actual: ["staged 2", "staged 3", "staged 1"]
  #expect(texts(delivered) == ["staged 1", "staged 2", "staged 3"])
}

/// #304's "re-scoped to a different topic in between".
@Test
func rescopingAWatchDropsTheWakesForTheTopicItLeft() async {
  let (store, target, sender) = await makeStore(readings: readings, delivered: delivered)
  await store.handle(.mailroomWatch(on: true, topic: "alpha", from: target))
  await store.handle(.mailroomPost(text: "an alpha post", topic: "alpha", from: sender))
  await store.handle(.mailroomWatch(on: true, topic: "beta", from: target))
  await readings.scriptTargetReads(Array(repeating: .idle, count: 20))
  await store.pollPresence()
  #expect(!delivered.value.contains { $0.contains("an alpha post") })   // fails
}
```

The other two — `aMessageStagedDuringADrainArrivesLastNotMerelyOnce` (asserts the full
order `[1, 2, 3, late]`, not just "once") and `watchOffLandingDuringADrainStillDropsTheWake`
(`--off` landing while the batch is already taken) — **pass on this head** and fail on
`main` with 12 deliveries for 4 messages and 4 for 2 respectively. They're worth adding to
`FollowUpDrainTests` as-is: the existing overlap test asserts `prefix(3)` and per-message
counts, which would not have caught a late message arriving *first*.

</details>

Issue #318, opened 2026-09-07
PresencePollingTests.aTickThatChangesNothingTellsNobody is flaky: probe.asked.count races the tick


scgopi:
Fails intermittently, ~1 in 5, independent of any recent change:

```
✘ Test aTickThatChangesNothingTellsNobody() recorded an issue at PresencePollingTests.swift:120:5: Expectation failed: await probe.asked.count == 1
✘ Test run with 1672 tests in 175 suites failed after 63.602 seconds with 1 issue.
```

Evidence (same machine, same head, 2026-09-06):

| Run | Result |
|---|---|
| Full suite, earlier | pass |
| Full suite, later | **fail** |
| Isolated ×5 | 1 fail / 4 pass |

The failure is an async count expectation racing the tick (failed after 0.001 s), not an assertion about state that changed. Unrelated to #317's diff (reaper/workspace/persistence); flagging it because a flake that fails a full-suite gate costs every loop gating behind one.

Suggested direction: make the probe's ask observable (a continuation/confirmation expectation) rather than polling `asked.count` immediately after the tick.

scgopi:
Closing as no longer reproducible. On the head of #321 (2fd34769, which carries every drain and presence change since this was filed), `PresencePollingTests` ran 30 iterations in isolation with `-test-iterations 30 -run-tests-until-failure`: 300/300 passed, `aTickThatChangesNothingTellsNobody` 30/30. The full 1692-test suite also passed on that head. No commit targeted this test specifically; if the count race comes back, reopen with the run.

PR #321, opened 2026-09-07
Retry a timed-out follow-up whose send then fails; fix the gate after #320

scgopi:
Makes `main` green after #320 and fixes the defect its exactly-once bookkeeping left behind. Two commits: the gate fixes the Release loop verified (`.claude/handoff/gate-fixes-320.patch`, applied as-is), then the store fix with tests.

## Gate fixes (commit 1)

- `DrainWedgeVerificationTests.swift` passed `deliveryDeadline:` last, but `GraphStore.init` declares it second, right after `graph:`. Swift matches an argument list in declaration order, so the test target did not compile; the reported `messageNode` / `refreshUsage` errors were downstream noise. Moved it to where it is declared.
- `ProjectPersistence.swift` failed `swift-format --strict` at 61:10, 62:1, 67:14. Reformatted with the formatter on that file only.

## The defect (commit 2)

#320 keeps a timed-out delivery on the queue until its abandoned send reports a verdict. A late **success** is removed, exactly once — correct. A late **failure** went through `staged(_:)`, which only writes the memory log and returns a marked copy; the return was discarded in both `finishTimedOutDelivery` and the drain's fold-back, so the item silently left the live queue. A target that was idle and answering never received text it was owed, while a send that failed *before* the deadline was retained for retry (the `!delivered` branch on main). Both failures are now treated the same: staged once, kept in place on the queue, and retried by a drain the verdict itself kicks — a failure learned this late has no command to ride the drain of, so without its own it waited for the next poll, or for good if none came.

### The test that was failing

`DeliveryWedgeTests.aHungDeliveryStillFreezesEveryOtherLoopsFollowUps` asserted `delivered` would contain the hung one's message, but its own `onDeliverMessage` closure never appended for the hung node — the hung branch returned `true` without recording. The last expectation was unreachable whatever the store did, and the store was in fact correct for that path: the send succeeded late, so the message must **not** be sent again. The closure now records its late success, and the test checks two further drains send nothing more. The new test `aTimedOutSendThatThenFailsIsRetriedExactlyOnce` covers the late-failure path and fails on main's `GraphStore`:

```
✘ aTimedOutSendThatThenFailsIsRetriedExactlyOnce … Expectation failed: (delivered.value → ["[graphcode] for the bystander"]) == ["[graphcode] for the bystander", "[graphcode] for the hung one"]
✘ … Expectation failed: (attempts.value → 1) == 2
```

Both tests now await the wedged command's return instead of a ten-second settle — the deadline is what returns it — and then refresh, because the bystander's message is queued mid-drain and moves on the pass after. The two suites run in 0.3s instead of 24s.

Not covered here: a send that never reports at all stays parked in `pendingDeliveryAttempts`, unstaged, until the daemon restarts. Same as #320's main; noted so it is not mistaken for fixed.

## Gate

In flight from `worktrees/fix-320-lost`, private DerivedData; numbers posted below when it finishes.

- `make check` exit 0 (swiftlint 0 errors, swift-format clean)
- targeted `DeliveryWedgeTests` + `DrainWedgeVerificationTests`: exit 0, 6 tests; red on main's `GraphStore.swift` (exit 65, the new test only)

https://claude.ai/code/session_014niBD1D1yQ7F2gjV2K54eB

### DIFF
--- a/GraphcodeKit/Sources/GraphStore.swift
+++ b/GraphcodeKit/Sources/GraphStore.swift
@@ -2884,7 +2884,7 @@ public actor GraphStore {
         var retained: [PendingFollowUp] = []
         for pending in remaining + Array(batch[index...]) {
           if let confirmed = completedTimedOutDeliveries.removeValue(forKey: pending.id) {
-            if !confirmed { _ = staged(pending) }
+            if !confirmed { retained.append(staged(pending)) }
           } else {
             retained.append(pending)
           }
@@ -2982,6 +2982,9 @@ public actor GraphStore {
         Task { [self] in
           let result = await attempt.wait()
           finishTimedOutDelivery(pending.id, confirmed: result)
+          // A failure learned this late has no command to ride the drain of; without
+          // its own it waits for the next poll to retry, or for good if none comes.
+          if !result { await drainPendingFollowUps() }
         }
         DaemonLog.shared.record(
           "delivery-stall",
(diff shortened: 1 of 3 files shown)


scgopi:
Gate on `2fd34769` (= remote head), from `worktrees/fix-320-lost`, private DerivedData:

| Step | Result |
|---|---|
| `xcodebuild test` (scheme graphcode) | exit 0 — 1692 tests / 180 suites / 0 failures; 0 `** TEST FAILED **`, 0 `Failing tests:` |
| `graphcode-cli` build | exit 0 |
| `graphcoded` build | exit 0 |
| `make check` | exit 0 — swiftlint 0 errors, swift-format clean |
| `scripts/cli-smoke.sh` | exit 0 on all five verbs, against a fresh `swift build` of this head (the worktree's `.build` link points at the main checkout's stale debug binary, so the script's own build was bypassed with a scratch path) |
| Linux build | pass — https://github.com/scgopi/GraphCode/actions/runs/34087301065/job/101633685121 |
| Red on main | `aTimedOutSendThatThenFailsIsRetriedExactlyOnce` exit 65 with main's `GraphStore.swift` (`attempts → 1`, hung message never re-sent); exit 0 with it restored |

Also, #318's `PresencePollingTests` ran 30 iterations in isolation on this head (`-test-iterations 30 -run-tests-until-failure`): 300/300 passed.

PR #342, opened 2026-09-12
fix: pi extension banks a session id only once its file exists

scgopi:
Refs #214. Found while reviewing #338.

## Problem

pi picks a session id at startup, but it only writes the session file when the first assistant message arrives. The presence extension banked the id at `session_start`, before that file existed. `pi --session <id>` exits 1 with `No session found matching '<id>'` when an id has no file. So the banked id was dead whenever:

- a pi session was quit before its first reply (e.g. a turn-based loop opened with no prompt), or
- a reboot interrupted a `/new` before its first reply. The dead id replaced the id of the conversation before it, so that conversation could no longer be resumed.

Locally the ensure and the pane notice the resume dying and fall back to a fresh launch. That fallback costs an extra launch and a `resume-dead` entry in the dial log, and it can't recover the conversation the dead id replaced.

## Fix

`bank()` now writes the id only once `getSessionFile()` exists on disk. It retries at `tool_call` and `agent_settled`, the first points where pi has saved an assistant message. A session that already has a file, such as a resumed one, still banks at `session_start`.

## Verified against pi 0.85.1

Tested with the real `pi` binary, a fake `zmx` shim, and the rendered extension:

| Launch | Before | After |
|---|---|---|
| Bare TUI, killed before any reply | id banked, no session file, `--session` exits 1 | nothing banked |
| TUI with a prompt (the provider error is saved as an assistant message) | banked, file exists | banked, file exists |

A new test, `anIDIsBankedOnlyOnceItsSessionFileExists`, checks that the file-existence check is in place and that both retry handlers call `bank()`.


scgopi:
Gate on `83131e3a` (= remote head), private DerivedData: `xcodebuild test` exit 0 — 1732 tests / 183 suites / 0 failures (+1 over the #338 base); `graphcode-cli` build exit 0; `graphcoded` build exit 0; swiftlint 0 errors; swift-format clean; cli-smoke exit 0; Linux pass.

Fails without the fix: with the base `PiPresenceExtension.swift` checked out, `anIDIsBankedOnlyOnceItsSessionFileExists` records 3 issues (xcodebuild exit 65).

PR #358, opened 2026-09-14
Add configurable Copilot runtime version preference

ZekeAranyLucas:
## Summary

Add an optional, version-agnostic Copilot runtime preference so GraphCode can launch a known-good CLI when a newer release regresses. No release is hardcoded: `1.0.84-5` is only an example.

## Changes

- Persist `copilotPreferredVersion` in the settings shared by the app and daemon; missing or blank values preserve normal Copilot version selection.
- Pass `--prefer-version <version>` through app and daemon launches and resumes, locally and over SSH, plus title and summary requests.
- Add **Preferred versions** immediately after **Permissions**, with a matching row for each backend. Copilot offers **Default** or **Specific version**, with a shell-quoted, copyable npm install command; Claude Code, Codex, OpenCode, and Pi remain disabled at **Default** until their version overrides are implemented.
- Keep backend selection unchanged: the version controls do not change the default backend for loops or Quick Chats.
- Document the required manual order: install on each execution host, verify the reported version, then activate the preference.
- Cover settings compatibility, launch/resume parity, other-backend isolation, headless requests, and shell argument handling.

## Scope and remaining rollback work

**This is version-preference plumbing, not a complete guided runtime rollback operation.**

GraphCode does not execute npm, verify installation readiness, or gate activation on a successful installation. Editing the preference takes effect immediately for subsequent launches. Existing sessions are not interrupted.

The intended complete flow is choose a version, explicitly approve installation, install and verify it on the execution host, and only then activate the preference. Global npm installation also changes Copilot outside GraphCode; that effect needs to be clear before approval. The guided install/verify/activate workflow remains outstanding.

## Test plan

- Shared library, daemon, and CLI: `swift build` passed.
- Initial app build and eight focused suites: 84 tests passed (CopilotVersionTests, GraphcodeSettingsTests, CopilotBackendTests, AttachedSessionPermissionTests, LocalSessionResumeTests, RemoteSessionResumeTests, SummaryStoreTests, TitleSuggestionTests). The SwiftPM build and all 84 tests also passed after rebasing onto `main`.
- After the Preferred versions UX update: the app built and 30 tests passed across CopilotVersionTests, GraphcodeSettingsTests, CopilotBackendTests, and AttachedSessionPermissionTests, including new coverage that version preferences start at Default and do not change backend selection.
- Strict Swift formatting passed for changed Swift files. Targeted SwiftLint passed for the new behavior; existing file/type-length violations in GhosttyTerminalView.swift and SummaryStoreTests.swift also reproduce on the baseline.
- Confirmed `copilot --prefer-version 1.0.84-5 --version` selects that installed version and npm publishes the matching package.

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] I have signed off my commits (`git commit -s`) per the DCO
- [ ] Full `make test` run (focused suites run instead, as listed above)
- [ ] Full `make check` run (targeted checks run; pre-existing length violations noted above)


scgopi:
Merging this. The two unticked boxes in your checklist are covered: we ran the full `make test` (1824 tests, 194 suites, all passing) and the full `make check` (exit 0) on the branch merged with current `main`, on macOS. The pre-existing length violations you noted in GhosttyTerminalView.swift and SummaryStoreTests.swift do reproduce on the baseline and are not from this change.

We also verified the cross-backend question directly, since it was the main risk: `--prefer-version` is emitted from exactly one place in production code, behind a `supportsVersionPreference` guard that is Copilot-only, and argv goldens across five backends and every launch shape are byte-identical with the preference unset and set. Removing the guard makes those goldens fail, so the coverage is real.

One fix we will land separately on `main`: the README's install command uses `@github/copilot@v1.0.84-5`, which 404s — no published version is `v`-prefixed. `1.0.84-5` is correct, as your next line has it.

The guided install/verify/activate flow you describe as outstanding is a good follow-up. Thanks for this.

Issue #215, opened 2026-08-29
Goal loop dies silently on Claude Code's 'trust this folder' prompt; stays idle and cannot be woken


Yuncun:
**Version**: graphcode.app 0.1.51 (macOS 26.4, Claude Code backend)

**What happened**
Created a goal loop from the CLI on a folder Claude Code had never been opened in:

```
graphcode node create /Users/ericshen/Claude/twodrive --title "TwoDrive backlog" --type goal --goal "..." --predicate .../tools/backlog-done.sh --metric .../tools/backlog-open-count.sh --direction minimize --skip-unchanged --backend claudeCode --model capable
```

`status` showed `running` for ~60 s, then `idle`. The zmx session had exited with code 1. `zmx history` shows Claude Code stopped on its first-run dialog:

```
Quick safety check: Is this a project you created or one you trust? ...
 ❯ No, exit
   Yes, I trust this folder
ZMX_TASK_COMPLETED:1
```

Nothing in the app or `graphcode status` says the session exited; the loop just looks idle. `graphcode node send <id> ...` reports `delivered` but nothing respawns (after `zmx kill` it reports "delivery failed — staged to memory", still no respawn). The only way back was `node delete` + `node create`.

**Expected**
1. A backend session that exits non-zero before its first turn should be surfaced (status line like `session exited (1)`), not shown as `idle`.
2. `node send` (or a `node start`/`resume` verb) should respawn a dead session.
3. Ideally the launcher handles the trust dialog for folders the user explicitly added — e.g. by pre-setting `projects[<path>].hasTrustDialogAccepted = true` in `~/.claude.json` before launching `claude`, which is what fixed it here.

**Workaround**: set `hasTrustDialogAccepted: true` for the project path in `~/.claude.json`, then delete and recreate the node.

PR #362, opened 2026-09-16
Tell the truth about a blocked loop, in the card and in the rollup

scgopi:
Found from a live card: a sketch loop sitting behind an unfired hand-off, its agent running commands in the pane, showing **BLOCKED** — and appearing in no attention queue. Two separate defects, one symptom.

## 1. A working loop read BLOCKED

`LoopNode.displayState` corrected only `.running`, on the reasoning that a blocked node is waiting on its edge whether or not its session is alive. That is true of a parked loop and false of this one:

- `opensOnHumanTap` **deliberately** lets a human open an attended blocked loop — its own doc calls the tap "the authorization the hand-off was guarding".
- An unattended child's session starts *before* the follow-up hand-off edge marks it blocked.

In both cases an agent is genuinely working behind a card that says it is stopped, and nothing anywhere moves the node out of `.blocked` — there is no `.blocked → .running` transition.

A live session now corrects the card: `busy` reads RUNNING, `awaitingInput` reads ASKING. Quiet, absent, unknown and a session that has exited all still read BLOCKED, because then the edge really is the whole story.

Raw `state` is untouched, exactly as for `.running` — edge firing, `MessageBus.deliverability` and the stranded rollup still read the node as blocked, because it is. This is the presentation of that fact, not a revision of it.

A side effect worth naming: an agent that asks a question from a blocked-but-live session now reaches the attention queue, which it could not before.

## 2. Permanently stranded loops were never surfaced

`AttentionRollup.strandedNodeIDs` asked whether an upstream could still deliver and answered `!source.isResolved`. That catches the upstream that failed and misses the one that simply stopped: a sketch only ever runs because a human opened it, so an idle one with no session will never resolve and therefore never fire. Every loop below it is stuck for good, while the rollup counts them as working as designed — the exact case the doc claims it reports.

`LoopNode.mayStillReachResolution` decides it on evidence instead: a live session, the launch grace, or a loop the daemon keeps alive. It also folds in #215's silent death — an unattended upstream whose session vanished is no longer treated as a pending hand-off.

Anything short of an actual presence reading still counts as hopeful. A missing or `.unknown` presence observed nothing, and Copilot and Codex report none at all; a guess that strands a healthy loop would put false alarms in the one queue whose entire value is that everything in it is real.

## Verification

| Check | Result |
|---|---|
| `xcodebuild -scheme graphcode … test` | exit 0 — **1839 tests in 195 suites** (was 1832; 7 added) |
| `-scheme graphcode-cli` / `-scheme graphcoded` build | exit 0 / exit 0 |
| `swiftlint lint` | 0 errors |
| Tests proved against `main` | reverting only the two source files → exit 65, both new behaviours fail by name |

The four guard tests (grace window, live session, no reading, blocked-stays-blocked) pass against `main` *and* this branch by design — they pin behaviour that must not change.

`swift format lint` cannot run on this machine at all: Xcode 27's toolchain rejects the checked-in `.swift-format` with `missing key orderedImports.shouldGroupImports`. Pre-existing and unrelated to this change.


scgopi:
Review from the `BlockedStateReview` loop (codex) came back CHANGES-NEEDED with one finding, confirmed and fixed in 5945cb5c.

**Duplicate rollup rows.** Correcting a blocked card to `.awaitingInput` put a node through `reason(for:)` that had never reached it before (a `.blocked` node previously always returned nil there). A stranded loop whose session is asking something was therefore appended twice — once by `items(across:)`, once by the stranded pass in `fullRollup`. Since `AttentionItem` is `Identifiable` on `nodeID`, that is a `ForEach` with a duplicate id before it is a confusing queue.

The stranded pass now skips a node already reported. `.blocked` ranks last in `AttentionReason` by design, so keeping the existing row always keeps the worse of the two reasons.

Regression test `aStrandedLoopThatIsAlsoAskingSomethingGetsOneRow` reproduces it: with only `AttentionRollup.swift` reverted it fails on `rollup.count == 1` (exit 65), and passes with the fix.

Re-verified: full suite exit 0, **1840 tests in 195 suites** (was 1839), `swiftlint lint` 0 errors.

PR #396, opened 2026-09-20
Let a failed remote delivery leave a trace on the host

scgopi:
Follow-up to #395. That fixed the broken installer; this fixes the reason nobody noticed for five days.

## The decision worth separating

The delivery fragment ended `>/dev/null 2>&1 || true`. Those are two different decisions wearing the same clothes:

- `|| true` — **keep**. A session without its briefing is still a session, and delivery must never block the launch it precedes.
- `>/dev/null 2>&1` — **wrong**. When the embedded python raised `SyntaxError` on every dial, the reason went to `/dev/null`. Neither machine held a word about why every remote host was empty.

The bug was never that the install could fail. It was that failure left no trace anywhere.

## Change

Failure stays non-fatal, and now reports to the host's own dial log beside the launch decisions it belongs next to:

```
2026-09-20T21:38:23Z delivery install failed NotADirectoryError: [Errno 20] Not a directory: ...
```

- Flattened to one line — every reader of `dials.log` splits on newlines.
- Bounded at 400 bytes, and taken from the **tail**. A python traceback opens with frames and interpreter paths and ends with the line that names the fault. Keeping the head threw the answer away: measured, `NotADirectoryError` fell outside the first 400 bytes of the very failure this was written to explain.
- The detail rides as a `printf` **argument**, not inside the format — unlike `session`/`dial`/`event`, which are literals this codebase controls. A `%s` in a remote python's error must not be able to reformat the line it is written to.
- A succeeding delivery writes nothing. The sweep dials every host every minute; a line per healthy tick would bury the one that isn't.

## One existing test changed

`aFailedDeliveryLeavesNoStampBehind` asserted `!script.contains("printf")` as a proxy for "the shell must not write the stamp". A `printf` now legitimately appears, so it asserts the rule itself: every `printf` in the fragment targets `dials.log`. The stamp assertions around it are untouched.

## TDD evidence

RED: xcodebuild -only-testing:graphcodeTests/RemoteInstallerExecutionTests with main's RemoteGraphAccess.swift and DialLog.swift in place -> exit=65, aFailedDeliveryLeavesItsReasonInTheHostsDialLog fails because no dials.log exists at all

GREEN: xcodebuild -scheme graphcode -destination platform=macOS test -> exit=0, ** TEST SUCCEEDED **, Test run with 1929 tests in 203 suites passed

REGRESSION: full suite plus graphcode-cli and graphcoded builds, swiftlint and swift format lint -> exit=0 on every step, 0 lint errors, the 1927 tests from the 0.1.74-beta1 tree all still pass


scgopi:
## Review findings addressed — `80a088a9`

`ReviewPR396` reviewed this adversarially and found five defects, four reproduced end to end. I independently reproduced the two HIGH ones before acting. All five are fixed.

| # | Finding | Fix |
|---|---|---|
| F1 HIGH | `tail -c 400` splits UTF-8 mid-character; one orphan byte makes `grep`/`sed` fail on the **whole** log | cut piped through `iconv -c -f UTF-8 -t UTF-8` |
| F2 HIGH | 445-byte lines vs a 209-byte per-line budget — `keptLines` × 445 = 2.2 MB against a 1 MB cap, so the trim can never get under it again | detail size **derived** from `maxBytes / keptLines`, locked by `DialLogBoundTests` |
| F3 MED | fixed `dials.log.tmp` name — a race this PR made fire on every append; 40 concurrent fragments left 34 of 5000 lines | per-process `$$` scratch name, trim factored into one place |
| F4 MED | non-zero exit with empty stderr logged no reason and dropped the exit code | falls back to `rc=<code>` |
| F5 TEST | my replacement assertion was a loose proxy that accepted what the old line forbade | asserts the rule directly: no redirect names the stamp, every append targets the dial log |

F1 deserves emphasis: it would have made this change **net-negative**. A feature written to make one failure legible would have made every other dial entry on that host unfindable. Confirmed here — `grep -c "ensure fresh"` returned nothing on a log that visibly contained the line.

F5 is mine to own. The previous commit's body claimed that assertion "asserts the rule itself". It did not, and the counterexample is a rogue `printf 'stamp' > …/.shim-stamp;` passing because the text after it picks up `dials.log` from the next fragment's trim.

The review also cleared five properties by running them: no variable collision with the ensure's own `gc_rc`/`gc_boot`/`gc_last`/`GRAPHCODE_RESUME_ID`; the exit-status contract holds through the full `exec zsh -l -i -c` round trip and inside `{ install || { log; false; }; } && launch`; `printf` format escape is impossible; no file contents reach the log; and a succeeding delivery writes nothing.

Also hardened without a finding: `\r` is now stripped, since `tr` handled only `\n` and `\t`.

### Verification

| Step | Result |
|---|---|
| `xcodebuild -scheme graphcode test` | `exit=0`, `** TEST SUCCEEDED **`, `Test run with 1935 tests in 204 suites passed` |
| `graphcode-cli` / `graphcoded` builds | `exit=0` each |
| `swiftlint` / `swift format lint --strict` | 0 errors / `exit=0` |

This PR is deliberately **not** going into the `0.1.74` stable. Nothing here is user-visible; the fix users are waiting on is #395, already in `0.1.74-beta1`. It rides the release after.

